PR #6, opened 2016-12-06
Add a way to proxy module imports to a local repo

jpellizzari:
Implements #4.

The README contains instructions on how to implement in a test project.

* The resolver module is written in a very generic way, so it could be used for other stuff.
* Hot reloading should work. If you save a file the `ui-components` dir, webpack in the project consuming the lib should do a reload as if its in the same directory.

Issue #5, opened 2016-12-06
Implement partial imports


jpellizzari:
It would be nice to be able to do `import SomeComponent from 'weaveworks-ui-components/SomeComponent'`, instead of importing the entire module. This will allow webpack to do tree-shaking and keep bundle sizes smaller.

jpellizzari:
This will not be possible with out-of-the-box imports with webpack. Other modules use some sort of publish step that formats the directory structure of the module. In order to do partial imports, there needs to be a directory for each sub module in the top level of the package. Here's an `ls -1` of  `material-ui` from node_modules; `lodash` follows a similar pattern:
```bash
...
RaisedButton
RefreshIndicator
SelectField
Slider
Snackbar
Stepper
Subheader
SvgIcon
Table
Tabs
TextField
TimePicker
Toggle
Toolbar
index.js
...
```

A custom resolver will need to be added in order for this to work.

jpellizzari:
Closed by #7

PR #7, opened 2016-12-07
Enable partial imports

jpellizzari:
Implements #5 . This is branched from #6  and should only be merged after it.

This resolver  will allow us to do `import Button from 'weaveworks-ui-components/Button` and let webpack to do tree-shaking to keep bundle sizes smaller.

From the README:
```javascript
const WeaveworksPartialImport = require('weaveworks-ui-components/resolvers').WeaveworksPartialImport;
module.exports = {
  ...
  plugins: [
    ...
    new webpack.ResolverPlugin(new WeaveworksPartialImport())
  ]
}
```

Since the `weaveworks-ui-components` package is hosted on github (and not `npm`), the package structure has to roughly map to the repo structure. Packages like `material-ui` and `lodash` have the luxury of a `publish` step in which the package structure is changed before being pushed to NPM. Therefore, we need this custom webpack resolver.


davkal:
Can haz rebase?

jpellizzari:
@davkal fixed

PR #13, opened 2017-01-17
Adds a MatchedText component!

foot:
- Designed initially for use with flux-ui service filtering


foot:
@jpellizzari PTAL!

src/components/MatchedText/MatchedText.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+
+    return res;
+  }
+  return chunks;
+}
+
+
+const STYLES = {
```
jpellizzari: Assigned and never used. Looks like CI isn't catching lint stuff. I will fix on master.

PR #16, opened 2017-01-26
add logoprogress and circularprogress spinners

bowenli:


jpellizzari:
Looks awesome! Only feedback would be to get rid of the material-ui blue on the circular progress in favor of our own blue from the logo: `rgb(0, 180, 237)`

jpellizzari:
Looks awesome! My only feedback would be to change from the material-ui blue on the CircularProgress to our own blue from the logo: `rgb(0, 180, 237)`

bowenli:
updated

PR #17, opened 2017-02-07
Adds disabled and styles props to the Button cmp

foot:


src/components/Button/example.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import Button from '.';
+
+export default function MatchedTextExample() {
```
jpellizzari: Class name copy/paste typo

src/components/Button/example.js
```diff
@@ -0,0 +1,17 @@
...
+  const onClick = () => console.log('clickity clack');
```
jpellizzari: Every example component gets a `clickHandler` prop than can be used to show off what the Component does. So here you can do:
```
const onClick = () => this.props.clickHandler('onClick', 'clickity clack');
```
And a dialog will pop up with the args you pass in. See the Menu/example.js for a reference. I guess I should document this somewhere...

Also I just fixed a bug on master where the dialog overlay doesn't appear. If you rebase, it should work.

PR #18, opened 2017-02-13
Adds a ListItem component for putting things in lists

foot:
Often a title + a subtitle

### DIFF
--- a/src/components/ListItem/example.js
+++ b/src/components/ListItem/example.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import { ListItem } from '.';
+
+
+export default function ListItemExample({ clickHandler }) {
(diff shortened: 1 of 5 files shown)


src/components/ListItem/example.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { ListItem } from '.';
+
+
+export default function ListItemExample({ clickHandler }) {
+  const onClick = () => clickHandler('onClick', 'clicked it');
```
jpellizzari: The idea behind the `clickHandler` was to use it directly on the example components, so we can see what they return to the callback, if anything.

jpellizzari:
@foot LGTM

PR #36, opened 2017-08-24
Use styled-components for button

aaron7:
Proposal to use styled-components instead of global sass.

Advantages: no global css (less chance of conflict, no need to use something like BEM), use less nested styles, components are now independent "modules", easier to test styling with snapshots

Disadvantages: CSS in JS is still a hot debate. Lots of discussion around this.

---

- `Button` works identically with all of the styles now contained within the component. (Although ThemeProvider is required).
- A theme file is used to define colours, font sizes etc and component specific configuration.
- `.attrs()` is used for `StyledButton` to create a prop called `nature` which is used to lookup button specific configuration in a neat way.
- Stateless components are used.

The future:
- All components would have independent styles, however still sharing themes, animations, mixins etc.
- All font sizes, margin sizes, widths, heights etc will be defined in the theme file explicitly. These will be the only sizes components can use.

Testing:
- Jest snapshots allows testing the rendering of components with different props easily.
- We only need to test additional logic such as custom onClick handlers. 
- We can avoid mount unless necessary. E.g. with Dialog we can still simulate the Button without rendering it (using shallow rendering).
- We can keep test cases rendering components only with the minimal information the test case needs. e.g. no need to render a component with action mocks if they are not being used for the test case. Components should be set up minimally for the particular test case. `beforeEach` is good, but is creating an object for each an every test case with a superset of props. If there are many common props, we could move them out to a const outside of the test cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,17 +24,21 @@
   "peerDependencies": {
     "font-awesome": "4.7.0",
     "lodash": "^4.17.2",
-    "react": "~15.5.0",
+    "react": "^15.6.1",
     "react-dom": "~15.5.0"
   },
   "dependencies": {
     "babel-cli": "^6.18.0",
     "babel-plugin-transform-export-extensions": "6.8.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-es2015": "6.18.0",
     "babel-preset-react": "6.16.0",
     "classnames": "^2.2.5",
     "node-sass": "4.5.3",
-    "prop-types": "^15.5.8"
+    "polished": "^1.4.1",
+    "prop-types": "^15.5.8",
+    "react": "^15.6.1",
+    "styled-components": "^2.1.2"
   },
   "devDependencies": {
     "babel-core": "6.18.2",
@@ -55,7 +59,8 @@
     "font-awesome-webpack": "0.0.4",
     "highlight.js": "9.8.0",
     "html-webpack-plugin": "2.24.1",
-    "jest": "17.0.3",
+    "jest": "20.0.4",
+    "jest-styled-components": "4.4.1",
     "json-loader": "0.5.4",
     "less": "2.7.2",
     "lodash": "4.17.2",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "lodash": "^4.17.2",
     "react": "^15.6.1",
     "react-dom": "~15.5.0",
-    "polished": "^1.4.1",
     "styled-components": "^2.1.2"
   },
   "dependencies": {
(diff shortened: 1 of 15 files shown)


src/components/Button/__snapshots__/Button.test.js.snap
```diff
@@ -0,0 +1,158 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
aaron7: This file is auto generated.

src/components/Button/Button.js
```diff
@@ -9,45 +44,11 @@ import classnames from 'classnames';
...
-        className={className || cl}
-        type={type}
-      >
-        {children || text}
-      </button>
-    );
-  }
-}
+const Button = ({ children, text, onClick, ...props }) => (
+  <StyledButton {...props} onClick={e => onClick(e, text)}>
```
jpellizzari: This will pass through invalid props to the `<button />` element. If I do `<Button selected />`, React will throw a warning about invalid props.

package.json
```diff
@@ -23,17 +23,21 @@
...
     "babel-plugin-transform-export-extensions": "6.8.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-es2015": "6.18.0",
     "babel-preset-react": "6.16.0",
     "classnames": "^2.2.5",
     "node-sass": "4.5.3",
-    "prop-types": "^15.5.8"
+    "polished": "^1.4.1",
+    "prop-types": "^15.5.8",
+    "react": "^15.6.1",
```
jpellizzari: This needs to be a peer dependency and a dev dependency

src/components/Button/Button.js
```diff
@@ -1,6 +1,41 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
+import styled from 'styled-components';
+
+const StyledButton = styled.button.attrs({
+  nature: (props) => {
```
jpellizzari: The key name `nature` confused me. Does `kind` make any more sense? `type` would make sense, but its already taken.

package.json
```diff
@@ -23,17 +23,21 @@
   "license": "UNLICENSED",
   "peerDependencies": {
     "lodash": "^4.17.2",
-    "react": "~15.5.0",
+    "react": "^15.6.1",
     "react-dom": "~15.5.0"
```
jpellizzari: Should `styled-components` also be a peer dependency, since service-ui uses it now?

jpellizzari:
@aaron7 Really nice work on this. My main concern is that we don't increase bundle size when consuming this lib by making sure dependencies are shared.

src/components/Button/Button.js
aaron7: styled-components filters out invalid props for us, but this is not ideal and can cause problems due to the large number of props they have to allow through. There is discussion here: https://github.com/styled-components/styled-components/issues/439 - passing props explicitly is something we could try. Will push a change in a moment.

src/components/Button/Button.js
aaron7: I will try using the name `type` (underneath the `style`)

PR #37, opened 2017-08-24
Add Input component

jpellizzari:
Adds an `<Input />` component with some basic validation features:

screen shot 2017-08-24 at 3 15 20 pm


bowenli:
If we can increase the space between the bottom of the label and the top of the input to 15px that would be good. Otherwise looks good.

Issue #38, opened 2017-08-29
Files without JSX are not transpiled by babel


jpellizzari:
Repro:

1. Add a new file that doesn't use JSX:
```
import styled from 'styled-components'

export const SomeComponent = styled.div` font-size { 24px};`;
```
2. Try to load the docs page. 
3. Fails with `Uncaught SyntaxError: Unexpected token import`
4. Add in some JSX to the file and it works fine.

This is probably due to dynamically building the page with hot reloading and tricky webpack context stuff (see `/docs/Root.js`). This would be the first time this was caught because all other components that are shown have JSX in them. 

Current workaround is to include JSX in the file.

PR #40, opened 2017-08-30
Text primitives

jpellizzari:
Relies on #36 

Adds a Text component to standardize fonts and font sizes  (not pictured is Extra Large {Bold, Italic}:
screen shot 2017-08-30 at 11 34 06 am

@bia I would like to get your input on the font-sizes I added to the theme:
```
+  fontSizes: {
+    normal: '1em',
+    large: '2em',
+    xl: '2.827em',
 },
```

Those roughly match what was discussed on the trello card.


src/utils/index.js
```diff
@@ -0,0 +1,17 @@
+// Working around https://github.com/weaveworks/ui-components/issues/38 by using commonjs imports
```
aaron7: It doesn't seem like #38 is affecting this. This works for me:

```
import { intersection, first, keys } from 'lodash';

// Return the value of a set of props against a lookup object.
// Helps when using "ternary" props, ie: <Text small bold />
// Third argument is the default value to return in the event of an `undefined`;
export const lookupValue = (props, lookup, _default) => {
  const matches = intersection(keys(props), keys(lookup));

  if (matches.length > 1) {
    throw new Error(`You have specified conflicting props: ${JSON.stringify(matches)}`);
  }

  return lookup[first(matches)] || _default;
};
```

aaron7:
Great :) lgtm

I may take a quick look at #38 if I get some time.

bia:
@jpellizzari looks good to me and matches the type-scale trello. I had originally imagined the h1 and h2 from that page (~4em, 2.827em), but now that I see your screenshot 2.827em, ~2em feels more appropriate. Did you ever try it with the ~4em xl size?

PR #45, opened 2017-09-08
Alert component

jpellizzari:
Adds an Alert component:
screen shot 2017-09-08 at 9 11 14 am

Also fixes some issues with `<Button />`:
- Button was ignoring the `style` prop that is used in a couple places in Weave Cloud
- Button would throw an error on click if no callback is provided. This props is optional.

### DIFF
--- a/src/components/Alert/Alert.js
+++ b/src/components/Alert/Alert.js
@@ -1,0 +1,73 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+
+import { fromAtoms } from '../../utils';
+
+const StyledAlert = styled.div`
+  background: ${fromAtoms('Alert', 'type', 'background')};
+  border-radius: 4px;
+  border: 2px solid ${fromAtoms('Alert', 'type', 'borderColor')};
+  color: ${fromAtoms('Alert', 'type', 'color')};
+  margin: 8px 0 8px 0;
+  min-height: 1em;
+  opacity: ${props => (props.visible ? '1' : '0')};
+  padding: 1em;
+  transition: opacity 0.2s linear;
+`;
+
+const CloseIcon = styled.i`
+  float: right;
+  cursor: pointer;
+  margin-left: 1em;
+`;
+
+/**
+ * An alert to let the user know about an action or state.
+ * ```javascript
+ * import { Alert } from 'weaveworks-ui-components';
+ *
+ * export default function() {
+ *  return (
+ *    <div>
+ *     <Alert type="warning" visible={true}>
+ *       Warning: Hmm, this is not good, but its not terrible.
+ *     </Alert>
+ *    </div>
+ *   )
+ * }
+ * ```
+ */
+function Alert(props) {
+  const { onClose, children } = props;
+  return (
+    <StyledAlert {...props}>
+      {children}
+      <CloseIcon onClick={onClose} className="fa fa-remove" />
+    </StyledAlert>
+  );
+}
+
+Alert.propTypes = {
+  /**
+   * Set the color scheme to indicate the nature of the alert.
+   */
+  type: PropTypes.oneOf(['default', 'success', 'warning', 'error']),
+
+  /**
+   * Toggle whether the alert is shown
+   */
+  visible: PropTypes.bool,
+  /**
+   * Callback that runs with the alert is dismissed by clicking the 'X'
+   */
+  onClose: PropTypes.func,
+};
+
+Alert.defaultProps = {
+  type: 'default',
+  visible: true,
+};
+
+
+export default Alert;
--- a/src/components/Alert/Alert.test.js
+++ b/src/components/Alert/Alert.test.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import renderer from 'react-test-renderer';
+import 'jest-styled-components';
+import { withTheme } from '../../utils';
+
+import Alert from './Alert';
+
+const c = renderer.create;
+
+describe('<Alert />', () => {
+  describe('snapshots', () => {
+    it('renders success', () => {
+      const tree = c(withTheme(<Alert type="success">Success</Alert>)).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+    it('renders warning', () => {
+      const tree = c(withTheme(<Alert type="warning">Warning</Alert>)).toJSON();
+      expect(tree).toMatchSnapshot();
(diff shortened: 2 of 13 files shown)


src/components/Alert/Alert.js
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+
+const fromTheme = (component, field) => props => props.theme.atoms[component][props.type][field];
```
aaron7: :+1: could we use something similar for Button as well?

src/components/Alert/Alert.js
```diff
@@ -0,0 +1,73 @@
...
+  border-radius: 4px;
+  border: 2px solid ${fromTheme('Alert', 'borderColor')};
+  color: ${fromTheme('Alert', 'color')};
+  margin: 8px 0 8px 0;
+  min-height: 1em;
+  opacity: ${props => (props.visible ? '1' : '0')};
+  padding: 1em;
+  transition: opacity 0.2s linear;
+
+  .fa-remove {
```
aaron7: it feels like we should avoid nesting and use:

```
const CloseIcon = styled.i`
  float: right;
  cursor: pointer;
  margin-left: 1em;
`
```

```
<CloseIcon onClick={onClose} className="fa fa-remove" />
```

but I'm unsure about conventions.

src/components/Alert/Alert.test.js
```diff
@@ -0,0 +1,36 @@
...
+      expect(tree).toMatchSnapshot();
+    });
+    it('renders default', () => {
+      const tree = c(withTheme(<Alert>Default</Alert>)).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+  });
+  it('closes when the "x" is clicked', () => {
+    const spy = jest.fn();
+    const alert = mount(withTheme(<Alert onClose={spy}>My alert</Alert>));
```
aaron7: should be able to use `shallow` instead of `mount`

jpellizzari: `shallow` might work now, but if more nested elements get added, it doesn't render them. `mount` ensures that this will always work when doing a `.find(selector)`.

aaron7: I believe `shallow` should always work with nested elements until you reach an external component. At that point, the test should probably be inside that component instead of it's parent.

We should always start with `shallow` and move up the chain only when required. This is usually when the test includes component lifecycles like `componentDidMount` or interacting with DOM APIs.

aaron7: We could try and follow this for most of the tests:
>Shallow rendering is useful to constrain yourself to testing a component as a unit, and to ensure that your tests aren't indirectly asserting on behavior of child components.

bowenli:
lgtm

src/components/Alert/Alert.test.js
jpellizzari: Why bother with `shallow` when `mount` will work no matter what? I shouldn't have to remember to change my test render function if I decide to change from a `<span>` to a `<Text />` for example. 

In this case, the `styled.i` is an external component, so shallow breaks:
```javascript
  it('closes when the "x" is clicked', () => {
    const spy = jest.fn();
    const alert = mount(withTheme(<Alert onClose={spy}>My alert</Alert>));
    alert.find('.fa-remove').simulate('click');
    expect(spy.mock.calls.length).toEqual(1);
  });
  it('closes when the "x" is clicked shallow', () => {
    const spy = jest.fn();
    const alert = shallow(withTheme(<Alert onClose={spy}>My alert</Alert>));
    alert.find('.fa-remove').simulate('click');
    expect(spy.mock.calls.length).toEqual(1);
  });
// Second one will fail
```

aaron7: > Why bother with shallow when mount will work no matter what?

I believe shallow is a lot faster which helps when you start to have 1000's of unit tests and still want them to run in a couple of seconds. A more useful reason to always start with `shallow` is to ensure you are writing unit tests just for the component.


> In this case, the styled.i is an external component, so shallow breaks:

`shallow` should actually still work in this case as it should render up to `<CloseIcon />` with the props (i.e. onClick) testable. I believe styled-components `withTheme` is causing an expected problem here. https://github.com/styled-components/styled-components/issues/624 has various workarounds.

`.dive()` to jump around the `withTheme` problem is one solution. Some created a function for this in the issue above.

```it('closes when the "x" is clicked shallow', () => {
    const spy = jest.fn();
    const alert = shallow(withTheme(<Alert onClose={spy}>My alert</Alert>)).dive();
    alert.find('.fa-remove').simulate('click');
    expect(spy.mock.calls.length).toEqual(1);
  });
```
To prove `withTheme` is causing a problem and `shallow` renders children, this also works (just removing the `withTheme`):
```it('closes when the "x" is clicked shallow', () => {
    const spy = jest.fn();
    const alert = shallow(<Alert onClose={spy}>My alert</Alert>);
    alert.find('.fa-remove').simulate('click');
    expect(spy.mock.calls.length).toEqual(1);
  });
```

I found https://github.com/airbnb/enzyme/issues/465#issuecomment-227697726 which looks good, but I think when it is mentioned that `shallow` does not render children, this is confused with not included at all. Instead it includes up to and including that level, just not further.

jpellizzari: Or I can avoid all of that and use `mount` :). I will take the performance hit in exchange for less false positives in my tests. 

It looks like using `dive` would require a reference to the child component in order to know what to look for.

In this case, the test fails with `shallow` erroneously, probably due to that `withTheme` issue you linked.

aaron7: > Or I can avoid all of that and use mount :). I will take the performance hit in exchange for less false positives in my tests.

I would like to use `shallow` where possible, but understand why you would like to use `mount`. It looks like we will have to disagree for this case.

Shallow rendering really helps with unit tests. It's not just rendering the DOM and calling lifecycle models that slows `mount` down, it's also because it renders all of the children down to the last child, which means the number of nodes in the tree increases exponentially as you go up each layer, which is a problem when you are testing higher up components (less so with these low level components). I think the difference in performance is an order of magnitude.

> It looks like using dive would require a reference to the child component in order to know what to look for.

All we would have to do is add `.dive()` with no arguments or use something like `export const shallowWithTheme = component => shallow(<ThemeProvider theme={theme}>{component}</ThemeProvider>).dive();`

I don't mind if `mount` is used for now - we can always revisit this later when it becomes a problem :)

PR #47, opened 2017-10-02
Change styleguide from markdown to react and add reference article

### LINKED ISSUE #39: Convert styleguide articles from markdown to react
Dropping a markdown file into the `/styleguide` directory will render it as a page. Instead, we would like to be able to drop a React component into that directory and have it render as a section of the style guide.

Requirements:
* Drop an exported react component into the `/styleguide` directory and have it show up
* Styleguide should be a single page, but sections should be navigable via anchors using the side navigation.

jpellizzari:
Fixes #39 

Allows for styleguide articles to be written as React components. This makes it easier to make more complex and useful pages.

@aaron7 PTAL at the implementation

@bowenli I would like to get your feedback on the content of the 'Colors' article I created. The goal is to provide some guidance on when to use which colors.


bowenli:
Ehhh.. Loose guidelines feel dangerous to me. I think green=success and red=fail is probably all we can safely say at this point.

jpellizzari:
@bowenli Can you elaborate please? What is dangerous about those guidelines?

bowenli:
@jpellizzari 
The style guide is meant to standardize on certain components when they are needed, not be a how-to for visual design. ie. if I need a button, I go to the style guide, I find the standard button. If I need to use orange, I go there and find the standard orange.

Saying something like "no more than 10% of the page should be accent colors" is trying to make this into an explanatory document for designing the page. I think if we're serious about revamping the style, then we should be tightening down decisions at this point to a few people (eg. you me bianca). I don't want to put out a 'how to' so other people can start playing around with colors, at least for now.

jpellizzari:
@bowenli Fair enough. The copy has been removed but the 'Colors' article is still there.

jpellizzari:
screen shot 2017-10-03 at 12 07 50 pm

jpellizzari:
Also changed 'Semantic' -> 'Status'

PR #54, opened 2017-10-18
Add Time Travel component

fbarl:
Continues on https://github.com/weaveworks/scope/pull/2888.

The component needs to live here if we're to reuse it outside of the Scope repo.

_Note to the reviewer:_ I had some difficulties testing this with proxying together with `scope` and `service-ui`, but things might work well after this PR is merged. If not, I'd put up a small follow-up PR to make sure the component can safely replace the one currently in Scope.


package.json
```diff
@@ -40,21 +40,25 @@
     "babel-preset-es2015": "6.18.0",
     "babel-preset-react": "6.16.0",
     "classnames": "^2.2.5",
+    "d3-drag": "1.2.1",
+    "d3-scale": "1.0.6",
+    "d3-selection": "1.1.0",
+    "moment": "2.19.1",
```
jpellizzari: Can we pin this to the patch version to avoid unwanted dupes? I'm worried that service-ui or scope will increment this and then increase bundle size by a lot. Another option would be to add it as a peer dep.

fbarl: Yes, that's a good point. I added this version to `peerDependencies` to lock it in across our frontend repos.

PR #56, opened 2017-10-26
Add earliest timestamp to time travel

fbarl:
Add support to _beginning of time_, i.e. the earliest timestamp we can jump to in Time Travel as required by https://github.com/weaveworks/scope/issues/2887.

##### Notes

* The new `earliestTimestamp` fields defaults to beginning of year 2014 if no value is provided.
* `onChange` callback has been renamed to `onChangeTimestamp`, so that needs to be updated in Scope when bumping the dep to `ui-components`.


fbarl:
[Time Travel earliest timestamp support](https://trello.com/c/6XBexwfa/165-time-travel-earliest-timestamp-support)

PR #59, opened 2017-11-03
Time Travel timestamps as strings, durations as numbers

### LINKED ISSUE #58: Time Travel timeline keeps jumping back when panning slowly
Probably has to do with the fact that the `timestamp` input param is a `moment()` object vs ISO string, so rendering gets triggered everytime a new object is sent, and not when the timestamp actually changes.

Making `timestamp` param a string would probably fix the problem.

fbarl:
Fixes #58.

Keeping timestamp arguments as plain strings removes the client app's dependency on `moment.js` and more importantly, enables React to do shallow comparison of `props` and be smarter about whether they've changed.

To make sure it works well in context of Weave Cloud, it might be best to test by proxying `service-ui` with both `scope` and `ui-components` repos, e.g.

`PROXY_SCOPE_COMPONENT=true SCOPE_PATH=/home/fbarl/Projects/weaveworks/scope/ PROXY_COMPONENTS=true COMPONENT_LIB_PATH=/home/fbarl/Projects/weaveworks/ui-components/ SERVICE_HOST=frontend.dev.weave.works yarn start`


fbarl:
[Time Travel timestamps as strings, durations as numbers](https://trello.com/c/ggcergno/183-time-travel-timestamps-as-strings-durations-as-numbers)

PR #62, opened 2017-11-09
Upgrade to React 16

aaron7:
Part of https://github.com/weaveworks/service-ui/issues/1210

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ See it in action:
 http://weaveworks-ui-components.s3-website-us-west-2.amazonaws.com/
 
 ## Prerequisites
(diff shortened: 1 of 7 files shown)


jpellizzari:
What about Scope?

fbarl:
Don't forget to update the README :) https://github.com/weaveworks/ui-components#prerequisites

aaron7:
> Change LGTM on its own, but Scope will be stuck on an old version until its upgraded. If we can avoid having duplicate react libs in our bundle, that would be good.

The plan is to upgrade scope and service-ui shortly after this :) We've checked the dependencies here: https://github.com/weaveworks/service-ui/issues/1210#issuecomment-343168440.

PR #74, opened 2017-11-28
Use normalize.css and fix button sizes

aaron7:
Add normalize to the component viewer with a font-family workaround which we are using in `service-ui`.

Fixed some button sizes which appear different on OSX and Chrome (corrected by normalize.css).

### DIFF
--- a/docs/js/main.js
+++ b/docs/js/main.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { ThemeProvider } from 'styled-components';
+import { ThemeProvider, injectGlobal } from 'styled-components';
+import styledNormalize from 'styled-normalize';
 
 import 'font-awesome-webpack';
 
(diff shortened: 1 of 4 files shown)


docs/js/main.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { ThemeProvider } from 'styled-components';
+import { ThemeProvider, injectGlobal } from 'styled-components';
+import styledNormalize from 'styled-normalize'
```
fbarl: Semicolon? That's why the tests are failing :)

docs/js/main.js
```diff
@@ -13,6 +14,19 @@ import Root from './Root';
...
 
+(() => injectGlobal`
+  ${styledNormalize}
+
+  /* https://github.com/necolas/normalize.css/issues/694 */
+  button,
+  input,
+  optgroup,
+  select,
+  textarea {
```
fbarl: Where did you take the list of elements from? why is it that e.g. `a` and `span` don't have to be listed?

aaron7: from the css linked from the issue

docs/js/main.js
```diff
@@ -13,6 +14,19 @@ import Root from './Root';
 import '../css/demo.scss';
 import '../img/favicon.ico';
 
+(() => injectGlobal`
+  ${styledNormalize}
```
fbarl: Just to make sure, this CSS applies only to the docs pages and won't be carried across to the users of these repo like `service-ui`, right?

aaron7: yes, it only applies to the docs page, however we will be using the workaround in `service-ui` as well

aaron7:
[Fix normalize.css layout bugs](https://trello.com/c/MMwEXmc7/206-fix-normalizecss-layout-bugs)

PR #76, opened 2017-11-28
Remove min-zoom constraint for Time Travel ranges

### LINKED ISSUE #75: Time Travel: remove min zoom constraint per range
Related to https://github.com/weaveworks/service-ui/issues/1508

fbarl:
Resolves #75.

##### Changes

* Removed zoom limit that depended on the selected range, making zoom min-zoom a constant again.
* Introduced a minimum width for the rendered intervals in Time Travel timeline to make selected range always visible.

### DIFF
--- a/src/components/TimeTravel/TimeTravel.js
+++ b/src/components/TimeTravel/TimeTravel.js
@@ -29,6 +29,7 @@ import {
   TICKS_ROW_SPACING,
   MAX_TICK_ROWS,
   TICK_SETTINGS_PER_PERIOD,
+  MIN_RANGE_INTERVAL_PX,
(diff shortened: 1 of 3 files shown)


fbarl:
[Time Travel: allow max zoom out with any range](https://trello.com/c/r5heEuMG/198-time-travel-allow-max-zoom-out-with-any-range)

fbarl:
> I guess some discussion can be had on the exact pixel number or even if it should be a measurement taking into account the PPI of the screen, say 2mm.

True, as you might have seen, there's quite a few constants like that in the Time Travel component and I chose all of them pretty arbitrarily, by what seemed to look good on my Chrome :)

PR #78, opened 2017-12-01
Split Time Travel into internal subcomponents

### LINKED ISSUE #70: Time Travel: Split the massive components into a bunch of private subcomponents
From #68:

> Time Travel component is becoming too massive (> 800 lines atm). We should distribute its logic between various subcomponents (RangeSelector already exists, we could make TimestampInput, TimeTravelStatus, Timeline, TimelineMarkers etc.. separate components).

Also, as @aaron7 has mentioned before, the components should be ideally be independent units, so we should consider bringing some of the util functions inside the component files and making them private to the component.

fbarl:
Resolves #70.

Added PropTypes to all the subcomponents and made a lot of them `React.PureComponent` to prevent excessive updating/rerendering.

##### Breaking Change

Time Travel component doesn't handle visibility anymore (through the animation triggered by `visible` prop), as it's technically doesn't have to do with the component logic. Scope will need to add that part to its `TimeTravelWrapper` component now when pulling in these changes.

##### New hierarchy of components

* TimeTravel
  * Timeline
    * TimelinePeriodLabels
      * TimelineLabel
    * TimelineRange
  * TimelinePanButton
  * LiveModeToggle
  * TimestampInput
  * RangeSelector


fbarl:
[Time Travel split code into subcomponents](https://trello.com/c/1y36UjFw/200-time-travel-split-code-into-subcomponents)

dlespiau:
FWIW, I don't find it shocking to have the visibility code inside ui-components. Ideally we want to make animation duration, easing, etc part of ui-components so we can unify the same way than the colour palette. Alternatively, animation constants could be defined here and used elsewhere.

fbarl:
Thanks for your review @dlespiau!

> I don't find it shocking to have the visibility code inside ui-components.

I had a couple of reasons why I thought it was better to get rid of it there actually:

* Removing any extra complexity from an already complex component
* Because of the CSS animation that was there for visibility, we had to explicitly set the component height for animation to work and that was specifically tailored for how we use it in Scope. Without the component controlling visibility, we can really treat it as an auto-sized block and decide how we want to embed it depending on the context we're in
* That sort of a pull down animation is likely to go away in Scope at some point anyway. I think it's more probable we'll only make the Timeline pullable while keeping the timestamp + controls there both in Explore and Monitor, but that's still a long way ahead

> Ideally we want to make animation duration, easing, etc part of ui-components so we can unify the same way than the colour palette. Alternatively, animation constants could be defined here and used elsewhere.

I agree, that was my initial idea with putting it into `ui-components` as well! Animation constants should also be part of our style guide :)

FYI I'm gonna postpone merging this until @aaron7's PRs are all merged and dev is working well again.

PR #79, opened 2017-12-01
Allow all minor versions for peer dependencies

aaron7:


package.json
```diff
@@ -34,12 +34,12 @@
...
   "peerDependencies": {
-    "font-awesome": "4.7.0",
-    "lodash": "^4.17.2",
-    "moment": "2.19.1",
+    "font-awesome": "^4.0.0",
+    "lodash": "^4.0.0",
+    "moment": "^2.0.0",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
-    "styled-components": "2.2.1"
```
fbarl: Hmm, I think I'd prefer us to use `~` here instead of `^` for those deps. I think we already had a case with something not functioning well after not having `moment` version locked, maybe it even happened for `font-awesome`.

p.s. Even if we lock the deps to the exact patch versions, things should be fine as we're versioning the `ui-components` repo so we do controled transitions from client repos, right?

aaron7: If we use `~` for these, we would need to release `ui-components` every time we would like to upgrade minor versions of these in `service-ui`.

Did the `moment` bug appear in `service-ui` or `ui-components`? It feels like the locking of packages such as moment should be in `service-ui` as long as the components in `ui-components` doesn't require a specific version of moment.

fbarl: Ok, following the Slack discussion, I can't find any good reasons for not relaxing the versions. I also can't remember any particular times it caused any harm, so I will approve :)

PR #80, opened 2017-12-07
Shave off two pixels from Timeline min-range width

fbarl:
Addresses https://github.com/weaveworks/service-ui/issues/1533#issuecomment-348170084 as part of https://github.com/weaveworks/service-ui/issues/1533.

Looks good on my Chrome and Firefox!


fbarl:
[Timeline interval too wide at high zoom level, and gap to center line](https://trello.com/c/1xKXxQt8/220-timeline-interval-too-wide-at-high-zoom-level-and-gap-to-center-line)

PR #81, opened 2017-12-08
Made Time Travel Timeline size-aware

fbarl:
Should finally resolve https://github.com/weaveworks/service-ui/issues/1533.

##### Changes

* Use `react-resize-aware` to have DOM triggers for resizing of timeline, not just the global viewport level ones.
* The DOM triggers still didn't help for a particular case of < > panning buttons icons being loaded, so fix the width of those buttons to `20px`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "polished": "^1.7.0",
     "prop-types": "^15.5.8",
     "react-motion": "0.5.2",
+    "react-resize-aware": "^2.7.0",
     "styled-normalize": "^2.2.1"
   },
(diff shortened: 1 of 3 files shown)


fbarl:
p.s. I've tested it on Firefox and Chrome (on Linux), what remains is OSX Safari :)

fbarl:
[Timeline interval too wide at high zoom level, and gap to center line](https://trello.com/c/1xKXxQt8/220-timeline-interval-too-wide-at-high-zoom-level-and-gap-to-center-line)

PR #83, opened 2017-12-11
Set up to publish to npm

aaron7:
Set up package.json and .npmignore to allow us to publish to npm. This allows us to not have to build (and install build dependencies) when installing `ui-components`.

- `yarn publish` to increment version and publish to npm
- `yarn docs:upload` to upload to s3 (requires aws account)

Tested locally installing the package and published to https://www.npmjs.com/package/weaveworks-ui-components. Will need to add our npm accounts as owners so others can publish.


jpellizzari:
The reason I didn't publish to NPM was because this lib isn't meant for public consumption. I also didn't want the hassle of dealing with NPM authentication. 

The trade off was building the files on install instead of before-hand on publish.  @aaron7 What problem is this solving?

aaron7:
> The reason I didn't publish to NPM was because this lib isn't meant for public consumption. I also didn't want the hassle of dealing with NPM authentication.
> 
> The trade off was building the files on install instead of before-hand on publish. @aaron7 What problem is this solving?

We are currently installing via git and extending `install` to build the library. This requires build dependencies to be included in `dependencies` instead of `devDependencies`. 

Therefore currently babel, node-sass etc are being included as dependencies of service-ui and scope: https://github.com/weaveworks/service-ui/blob/d725822fc4f6cd4eac40308be1d63d08caed819c/client/yarn.lock#L6993

jpellizzari:
@aaron7 To my knowledge, those don't end up in the bundled assets that get served to the user's browser. All this change would do is save us a little disk space and download time in our build container.

To me, the trade off for having those dependencies mis-categorized is worth the additional complexity of adding the NPM registry (auth, etc) into the mix. Thoughts?

aaron7:
_Discussing on slack_

> @aaron7 To my knowledge, those don't end up in the bundled assets that get served to the user's browser. All this change would do is save us a little disk space and download time in our build container.

I believe this is correct as it's not reachable via the entry point.

> To me, the trade off for having those dependencies mis-categorized is worth the additional complexity of adding the NPM registry (auth, etc) into the mix. Thoughts?

The overhead feels minimal (being logged into npm, which we both already are for other packages). Advantages would be:
- build on publish instead of building on every install (more immutable builds - although we do already lock packages). We could also achieve this by committing the `lib` however this is more likely to become out of sync.
- can keep deps & devDeps separate in ui-components
- can use npm versions instead of linking git repo tag

fbarl:
I don't have such a strong opinion on this, but with the assumption that we don't mind exposing our components on the NPM registry, I'm slightly in favor of @aaron7's arguments as build on publish and clear separation between deps & devDeps is something new contributors to the repo might appreciate (I feel that in a long run someone else is likely to try moving libs into devDeps like Aaron did).

So if I'm to break the tie, I'd say let's go with this :)

package.json
```diff
@@ -4,22 +4,25 @@
...
+    "build:components": "babel src/components --out-dir lib/components --ignore *.test.js,example.js",
+    "build:index": "babel src/index.js --out-file lib/index.js",
+
+    "release": "yarn ci && yarn build && npm version patch && git push origin --follow-tags",
+    "prepublish": "yarn ci && yarn build",
+
+    "docs:build": "RELEASE=true rm -rf dist && webpack --config webpack.config.js",
+    "docs:copy-images": "cp -R src/images dist/images",
+    "docs:s3": "aws s3 sync ./dist/ s3://weaveworks-ui-components --acl public-read",
+    "docs:upload": "yarn ci && yarn run docs:build && yarn docs:copy-images && yarn docs:s3",
```
fbarl: Not sure why e.g. `yarn docs:s3` works without `run` but then we have `yarn run docs:build`. Just to be on a safe side (and also for consistency), I think I'd do `yarn docs:*` -> `yarn run docs:*` and perhaps even `yarn ci` -> `yarn run ci`.

package.json
```diff
@@ -4,22 +4,25 @@
...
+
+    "build": "yarn build:theme && yarn build:utils && yarn build:constants && yarn build:components && yarn build:index",
+    "build:theme": "babel src/theme --out-dir lib/theme --ignore *.test.js",
+    "build:utils": "babel src/utils --out-dir lib/utils --ignore *.test.js",
+    "build:constants": "babel src/constants --out-dir lib/constants --ignore *.test.js",
+    "build:components": "babel src/components --out-dir lib/components --ignore *.test.js,example.js",
+    "build:index": "babel src/index.js --out-file lib/index.js",
+
+    "release": "yarn ci && yarn build && npm version patch && git push origin --follow-tags",
+    "prepublish": "yarn ci && yarn build",
```
fbarl: Could we somehow make sure this is always run before `npm publish`? That'd be a good safeguard I think..

package.json
```diff
@@ -4,22 +4,25 @@
...
+    "lint": "eslint src docs",
+    "test": "jest",
+
+    "build": "yarn build:theme && yarn build:utils && yarn build:constants && yarn build:components && yarn build:index",
+    "build:theme": "babel src/theme --out-dir lib/theme --ignore *.test.js",
+    "build:utils": "babel src/utils --out-dir lib/utils --ignore *.test.js",
+    "build:constants": "babel src/constants --out-dir lib/constants --ignore *.test.js",
+    "build:components": "babel src/components --out-dir lib/components --ignore *.test.js,example.js",
+    "build:index": "babel src/index.js --out-file lib/index.js",
+
```
fbarl: These empty lines get automatically removed when you try `npm install`, so maybe you can as well remove them now..

package.json
fbarl: Btw. why not include this in _relase_ or _publish_ flows? Is there a case when we'd want to e.g. release but not upload the docs immediately after?

fbarl:
Btw. not sure why `service-ui` tests are failing, maybe something here requires fixing: https://circleci.com/gh/weaveworks/service-ui/3199?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

package.json
aaron7: `prepublish` is a special script which always runs before `publish` (and local install)

package.json
aaron7: yarn doesn't require `run` to be used anymore. to me it feels more consistent to have `yarn test` `yarn build` `yarn randomscript` etc instead of some being prefixed with `run`. especially as scripts can all be overriden in package.json at the same level.

aaron7:
> Btw. not sure why service-ui tests are failing, maybe something here requires fixing: https://circleci.com/gh/weaveworks/service-ui/3199?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

These tests were failing due to the styles under src not appearing in the package. I've adjusted npmignore to include these, however we should ideally compile these at build time.

PR #84, opened 2017-12-12
Keep style files in npmignore

aaron7:
https://stackoverflow.com/a/5534865
> The pattern dir/ excludes a directory named dir and (implicitly) everything under it.
> With dir/, Git will never look at anything under dir, and thus will never apply any of the “un-exclude” patterns to anything under dir.

Currently the files inside the styles folder are not being included in the package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,7 @@
 bin/
 dist/
 docs/
-src/
+src/*
 !src/styles/
 styleguide/
 


dlespiau:
Can I ask why you need to keep the folder?

aaron7:
> Can I ask why you need to keep the folder?

`service-ui` imports `/styles.scss` which imports files from `/src/styles/*`. Ideally we would compile these in `ui-components` at build time, however for now we would like to continuing importing them as they appear in src.

PR #89, opened 2017-12-14
Make forward timeline pan button able to switch to live mode

### LINKED ISSUE #69: Time Travel: Auto-switch to live mode should also happen when clicking the right arrow
Introduced with the live mode support for Time Travel in #68.

fbarl:
Resolves #69.

##### Changes

* **Minor API change:** `onTimestampLabelClick` handler has been renamed to `onTimelineLabelClick`
* Added a new handler `onTimelinePanButtonClick` that is now distinguished from clicks on timeline timestamp labels
* Made forward timeline pan button sticky switch to live mode when moving to the current timestamp

### DIFF
--- a/src/components/TimeTravel/TimeTravel.js
+++ b/src/components/TimeTravel/TimeTravel.js
@@ -239,7 +239,7 @@ class TimeTravel extends React.Component {
       this.setState({ showingLive: false });
       this.setFocusedTimestamp(timestamp);
     }
-    this.props.onTimelineLabelClick();
+    this.props.onTimelinePanButtonClick();
   }
 
   handleTimelineZoom(duration) {
--- a/src/components/TimeTravel/TimeTravel.js
+++ b/src/components/TimeTravel/TimeTravel.js
@@ -102,8 +102,12 @@ function initialDurationMsPerTimelinePx(earliestTimestamp) {
  *      // track timestamp input edit...
  *    }
  *
- *    handleTimestampLabelClick() {
- *      // track timestamp label click...
+ *    handleTimelinePanButtonClick() {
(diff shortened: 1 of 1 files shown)


fbarl:
[Time Travel: Auto-switch to live mode should also happen when clicking the right arrow](https://trello.com/c/bykuZwPO/223-time-travel-auto-switch-to-live-mode-should-also-happen-when-clicking-the-right-arrow)

dlespiau:
Small reviewer note: when there are several independent changes like here, I find it easier to review if they are split across commits that I can fly through.

src/components/TimeTravel/TimeTravel.js
```diff
@@ -351,9 +363,13 @@ TimeTravel.propTypes = {
    */
   onTimestampInputEdit: PropTypes.func,
   /**
-   * Optional callback handling clicks on timeline timestamp labels (e.g. for tracking)
+   * Optional callback handling clicks on timeline pan buttons (e.g. for tracking)
+   */
+  onTimelinePanButtonClick: PropTypes.func,
```
dlespiau: This is never called anywhere.

fbarl: Good catch, I will modify the `handleTimelinePanButtonClick` function!

fbarl:
> Small reviewer note: when there are several independent changes like here, I find it easier to review if they are split across commits that I can fly through.

Gotcha, this time I just did it in one go and noticed it consisted of a couple of changes after commiting, but next time I'll try to split it better.

p.s. The example was broken a bit so I fixed it with one extra commit, will merge when the tests pass.

Issue #65, opened 2017-11-16
lightgray may be too light


dlespiau:
We've seen a number of badly calibrated monitors that don't really show `lightgray` https://github.com/weaveworks/ui-components/blob/master/src/theme/weave.js#L18 well and make it as white as white.

We may want to make lightgray a bit more gray

fbarl:
Since recently, @bia and I have been using http://snook.ca/technical/colour_contrast/colour.html for testing contrast compatibility between neighbouring colors - basically it tells us whether really bad screens still differentiate between two colors enough :)

bia:
Yes, as a quantitative guide we've been only choosing color combinations which are WCAG 2 AA Compliant.

PR #93, opened 2017-12-16
Change the shade of grey in use throughout the product

bia:
I'd like the background colour of the right side box of the monitor dashboards to match the left side navigation bar colour.

Please correct me if I'm wrong, but I think that we use fafafc far more than f8f8f8?

171212sidebarstandard1

cc @weaveworks/keyboard-orchestra

### DIFF
--- a/src/components/Alert/__snapshots__/Alert.test.js.snap
+++ b/src/components/Alert/__snapshots__/Alert.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`<Alert /> snapshots renders default 1`] = `
 .c0 {
-  background: #f8f8f8;
+  background: #fafafc;
   border-radius: 4px;
   border: 2px solid #aaaaaa;
   color: #000;
--- a/src/components/Button/__snapshots__/Button.test.js.snap
+++ b/src/components/Button/__snapshots__/Button.test.js.snap
@@ -18,7 +18,7 @@ exports[`<Button /> accepts a style prop 1`] = `
 .c0:hover {
   -webkit-transition: color .3s ease;
   transition: color .3s ease;
-  background: #f8f8f8;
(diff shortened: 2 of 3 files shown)


dlespiau:
Huum, we noticed that lightgray was already barely visible on some monitors, https://github.com/weaveworks/ui-components/issues/65 and this is making it even lighter.

bia:
@dlespiau Oh - good to know.
I'm just trying to keep things consistent. In that case we could keep #f8f8f8 and change the flux background, monitor side menu bar, and scope background to match. Does that sound ok?

dlespiau:
If possible, making it darker would be awesome. But otherwise, yes the darker of the two options is probably a good idea.

bia:
Even darker than #f8f8f8?
(let's talk about this in our sync today)

bia:
In the KO weekly meeting @dlespiau has agreed to help me to find the lightest grey possible to be differentiated from white on poorly calibrated monitors. Then I will reference that colour from the style guide throughout weavecloud. The assumption is that #f8f8f8 will be best, but I await @dlespiau's confirmation.

bia:
We're sticking with #f8f8f8, further changes will be requested in service-ui and scope.

bia:
>Huum, we noticed that lightgray was already barely visible on some monitors, #65 and this is making it even lighter.

@dlespiau This only matters if it needs to be differentiated from another color (implicit question : which color?)

rade:
@bia the lack of differentiation was between the background in the right pane of the service resource dashboard (e.g. https://cloud.weave.works/prom/cold-sky-72/dashboard/service-resources) and the general dashboard background.

bia:
@rade Right, trying to address exactly this in #167.

PR #99, opened 2017-12-21
Add click callback for disabled buttons

leth:
So that you can explain to the user why the button is disabled!


rndstr:
Isn't that sort of an anti pattern? If a user needs to click on a disabled button to be told why it's been disabled? Maybe a title/tooltip makes more sense

leth:
Yeah good point!

PR #103, opened 2018-01-04
Change Time Travel live mode toggle to icons

### LINKED ISSUE #71: Time Travel: small improvements to the UI
Two points from #68 are:

> * See if we settle at the current range intervals: 15min, 30min, 1h, 3h, 6h, 24h
> * Consider alternative wording to live / paused states (and perhaps tweak the styling around it)

cc @bia anything else that comes to your mind?

fbarl:
Resolves #71 by replacing the text with icons in Time Travel live mode toggler.

image
image

### DIFF
--- a/src/components/TimeTravel/TimeTravel.js
+++ b/src/components/TimeTravel/TimeTravel.js
@@ -32,6 +32,7 @@ const TimeControlsWrapper = styled.div`
 
 const TimeControlsContainer = styled.div`
   border: 1px solid ${props => props.theme.colors.neutral.lightgray};
+  box-shadow: 0 0 2px ${props => props.theme.colors.neutral.lightgray};
   background-color: ${props => props.theme.colors.white};
   border-radius: 5px;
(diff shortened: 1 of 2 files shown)


fbarl:
@bia I added a tiny commit which removes the pointer cursor for the pressed toggle button (which is currently made disabled i.e. not clickable). I was thinking that alternatively we could make both buttons always clickable and that whichever you click, the state is always toggled (as in, when you press the active button, it releases it like with some old tape players :)). Which option do you prefer?

fbarl:
> Yes! After trying it, I really like that idea. Then it feels like one larger button instead of two smaller ones.

Yeah, my only argument against this approach would be that it makes it a bit less clear from the icons which state we're in if both are clickable :/ But it's your call @bia, I really couldn't decide which one feels better :)

bia:
Yeah, I understand your point, but I still think that it would feel good if both were clickable. Let's go for it. :)

PR #108, opened 2018-01-05
Balance dropdown paddings

rndstr:
- Reduce left-padding from 24px to 12px
- Introduce padding at top/bottom of items container to balance ws

ui-components-dropdown

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -11,7 +11,7 @@ const Item = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-  padding: 0 24px;
+  padding: 0 12px;
   cursor: pointer;
 `;
 
@@ -25,6 +25,7 @@ const Popover = styled.div`
   margin-top: 4px;
   width: ${WIDTH};
   box-sizing: border-box;
+  padding: 6px 0;
 `;
 
 const Overlay = styled.div`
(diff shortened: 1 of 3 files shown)


aaron7:
_from slack discussion:_ the vertical centering of the selected item is very slightly too much to the top in the screenshot above.

We believe this may be only on linux as testing on OSX looks ok:
image

Do we think this is enough to create an issue?

rndstr:
Thanks for the addendum @aaron7 . I think we can leave it. Looking at other elements in other websites they are all rendered that way for me, must be an os/fontrenderer setting :)

My centering algo must include the descender (bottom hook of the `g`) and yours does so only by cap height which is much more pleasing to the eye, admittedly.

Can't unsee now :see_no_evil:

src/components/Dropdown/Dropdown.js
```diff
@@ -44,15 +45,6 @@ const ItemWrapper = Item.extend`
   &:hover {
     background-color: #eee;
   }
-
-  &:first-child {
```
jpellizzari: Why do these need to be changed?

src/components/Dropdown/Dropdown.js
```diff
@@ -11,7 +11,7 @@ const Item = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-  padding: 0 24px;
```
jpellizzari: I prefer 24. This also matches the `<Input />` component.

jpellizzari: The consensus in Slack seems to be less LR padding. Please make the `<Input />` match. I probably should have just set that in the theme somewhere.

src/components/Dropdown/Dropdown.js
rndstr: Because of https://github.com/weaveworks/ui-components/pull/108/files#diff-dae838509391e35fda596d1211449cdbR28, the background of a selected item no longer touches the rounded corner

PR #113, opened 2018-01-11
Allow inlining and centering of CircularProgress

rndstr:
- Add `inline` property
- Add `center` property

### DIFF
--- a/src/components/CircularProgress/CircularProgress.js
+++ b/src/components/CircularProgress/CircularProgress.js
@@ -9,17 +9,28 @@ import PropTypes from 'prop-types';
  * import { CircularProgress } from 'weaveworks-ui-components'
  *
  * React.render(
- *   <CircularProgress size={45} />
+ *   <div>
+ *     <CircularProgress size={35} />
+ *     <CircularProgress inline /> Inline
+ *     <CircularProgress center />
+ *   </div>
  * );
  * ```
  */
 class CircularProgress extends React.Component {
   render() {
-    const {size} = this.props;
+    const {size, inline, center} = this.props;
     const styles = {
       width: size,
       height: size
     };
+    if (inline) {
+      styles.display = 'inline-block';
+      styles.verticalAlign = 'middle';
+    }
+    if (center) {
+      styles.margin = '0 auto';
+    }
 
     return (
       <div className="weave-circular-progress-wrapper" style={styles}>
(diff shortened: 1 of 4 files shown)


src/components/CircularProgress/CircularProgress.js
```diff
@@ -9,17 +9,28 @@ import PropTypes from 'prop-types';
...
     const styles = {
       width: size,
       height: size
     };
+    if (inline) {
+      styles.display = 'inline-block';
+      styles.verticalAlign = 'middle';
+    }
+    if (center) {
+      styles.margin = '0 auto';
```
fbarl: Maybe it'd set this rather as `marginLeft: auto` and `marginRight: auto`, not to impose anything on the vertical margins.

src/components/CircularProgress/CircularProgress.js
```diff
@@ -9,17 +9,28 @@ import PropTypes from 'prop-types';
  * import { CircularProgress } from 'weaveworks-ui-components'
  *
  * React.render(
- *   <CircularProgress size={45} />
+ *   <div>
+ *     <CircularProgress size={35} />
+ *     <CircularProgress inline /> Inline
+ *     <CircularProgress center />
+ *   </div>
```
fbarl: The example looks tiny bit like something's broken:

image

Maybe adding some paddings/explanations in between would help (e.g. like in a Time Travel example).

src/components/CircularProgress/CircularProgress.js
```diff
@@ -9,17 +9,28 @@ import PropTypes from 'prop-types';
...
+    if (inline) {
+      styles.display = 'inline-block';
+      styles.verticalAlign = 'middle';
+    }
+    if (center) {
+      styles.margin = '0 auto';
+    }
 
     return (
       <div className="weave-circular-progress-wrapper" style={styles}>
```
fbarl: I think it would be preferable to do this by creating a new `CircularProgress` styled component and then forwarding `inline` & `center` to it as props. We're generally trying to move towards styled components whenever possible and `ui-components` repo is a good example where it would work nicely. Maybe you can even remove the `className` here and then update the styles in the PR that bumps the version in `service-ui`.

rndstr: Agreed, but I'd rather this to be done in a separate PR.

src/components/CircularProgress/CircularProgress.js
rndstr: Totally agree. Created reusable components and slightly modified them (left-aligned headings).

rndstr:
@fbarl PTAL

PR #116, opened 2018-01-11
Implement `leftIcon` property for `ListItem`

rndstr:
We have this in use in at least one place (as a leftover from the `material-ui/List` days)

### DIFF
--- a/src/styles/_components.scss
+++ b/src/styles/_components.scss
@@ -150,12 +150,19 @@
 
 
 .weave-list-item {
-  display: block;
+  display: flex;
   position: relative;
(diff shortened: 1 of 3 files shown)


src/styles/_components.scss
```diff
@@ -150,13 +150,20 @@
 
 
 .weave-list-item {
```
fbarl: While you're at it, maybe you could also make this chunk a styled component? :)

rndstr: :see_no_evil: might tackle this in a separate PR!

Issue #49, opened 2017-10-10
Alert: Add a property to control the visibility of the close button


dlespiau:
We have some cases where we currently don't provide the option to close the "Alert" (GKE instructions, flux errors in the setup page). It'd be nice if `Alert` could have a property to hide the cross.

PR #117, opened 2018-01-11
Only display close icon if click handler is given

rndstr:


jpellizzari:
Fixes https://github.com/weaveworks/ui-components/issues/49

PR #120, opened 2018-01-11
Make list-item content width 100%

jpellizzari:
Fixes https://github.com/weaveworks/service-ui/issues/1760


rndstr:
Thanks for the fix!

PR #132, opened 2018-02-14
Show time in UTC format in PrometheusGraph tooltip

fbarl:
Fixes https://github.com/weaveworks/service-ui/issues/743.

The timestamp format is uglier (less humanized), but at least it's consistent with how Time Travel control shows it. We should probably decide on a better cross-app time format, but that's a separate issue (cc @bia).

image


aaron7:
Related to https://github.com/weaveworks/service-ui/issues/743

PR #140, opened 2018-03-02
Smarter default PrometheusGraph series naming

fbarl:
`getSeriesName` prop now defaults to the following set of naming policies (in that order):

* If series has no metrics, name it as query string
* If series has **only one** metric, name it as the value of that metric
* If series has `__name__` key, throw the name in front and stringify the rest as JSON
* Otherwise simply stringify the whole metrics JSON

The only visible change is the second point (see the legend in the screenshot below), which fixes https://github.com/weaveworks/service-ui/issues/1886 (@rade can be the judge if enough to close that issue) and makes some preparatory work for https://github.com/weaveworks/service-ui/pull/2047.

image

### DIFF
--- a/src/components/PrometheusGraph/PrometheusGraph.js
+++ b/src/components/PrometheusGraph/PrometheusGraph.js
@@ -14,6 +14,10 @@ import {
   fromPairs,
   zipObject,
   forEach,
+  omit,
+  size,
+  has,
+  get,
 } from 'lodash';
 import { scaleLinear, scaleQuantize } from 'd3-scale';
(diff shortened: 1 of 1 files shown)


rade:
weaveworks/service-ui#1886 concerns what gets shown on hover, which the above screenshot doesn't show.

fbarl:
> weaveworks/service-ui#1886 concerns what gets shown on hover, which the above screenshot doesn't show

Right, I've updated the screenshot.

rade:
in which case I can confirm that this fixes weaveworks/service-ui#1886 :)

PR #147, opened 2018-03-12
Add time metric unit to PrometheusGraph

### LINKED ISSUE #146: PrometheusGraph - support time metric units
Prerequisite for https://github.com/weaveworks/service-ui/issues/2092.

fbarl:
Resolves #146.

##### Changes

* Added `seconds` metric unit type to support time units
* Changed `none` to `numeric` metric type for consistency with backend

image

### DIFF
--- a/src/components/PrometheusGraph/PrometheusGraph.js
+++ b/src/components/PrometheusGraph/PrometheusGraph.js
@@ -127,7 +127,7 @@ const colorThemes = {
 };
 
 const valueFormatters = {
-  none: (number) => {
+  numeric: (number) => {
     const step = number / 7;
     const formatNumber = number > 10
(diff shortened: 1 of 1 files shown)


dlespiau:
I would have found it more natural to read 1.4ms (or 1.391ms)

fbarl:
> Is there a good JS humanizer library we could use instead of doing this ourselves?

There are a couple (like `moment` & `d3-time-format`), but I didn't find any that would do what we want and require less lines of code to configure right :)

> I would have found it more natural to read 1.4ms (or 1.391ms), just like 1.2s is more natural than 1200ms.

Me too. I tried playing with it a little, but it seems slightly non-trivial to tweak it right, so I'm gonna give up this time and leave it for another issue, as it's more important to get some sort of support for time formats working for dashboards.

Merging...

Issue #161, opened 2018-03-28
Collapse cyan variants of theme colors


fbarl:
Usage across Weave Cloud:

![](https://placehold.it/15/00d2ff/000000?text=+) _accent.blue_, _weaveblue_
  * A lot of places, should probably keep as is since it's a brand color

![](https://placehold.it/15/00bcd4/000000?text=+) _turquoise_
  * Billing usage bar border
  * Background for button to start katacoda tutorials
  * Primary button background

![](https://placehold.it/15/39cbde/000000?text=+) _scooterLight_
  * Hover background for button to start katacoda tutorials (should turn that to ui-components button)

![](https://placehold.it/15/2db5ca/000000?text=+) _scooter_
  * Service error page form border

![](https://placehold.it/15/99edff/000000?text=+) _anakiwa_
  * Scope graph node label text match background (should probably be weaveblue with decreased opacity)

![](https://placehold.it/15/00a8cc/000000?text=+) _cerulean_
  * Scope matched results 'more'
  * Code editor dropdown text highlight
  * Instance deleted info check icon

---

I suggest we proceed as following:

* `accent.blue` should be kept as a brand color, but `weaveblue` can be collapsed into it directly and `anakiwa` can be made a semitransparent variant of it
* If we use the Button component for Katacoda tutorials, we can get rid of `scooterLight`
* `scooter` can be merged into `turquoise`
* `cerulean` can be split between `turquoise` and `status.info`
* This way, we will end up with only `accent.blue` and `turquoise` as cyan shades

bia:
This might be better dealt with in a separate issue, but could we shift turquoise to something a little darker, like #037AA9?
This colour passes WCAG AA both as a background with white text and as colored text on a #fafafc background

fbarl:
> This might be better dealt with in a separate issue, but could we shift turquoise to something a little darker, like #037AA9?

I'm ok making a small PR to change turquoise ![](https://placehold.it/15/00bcd4/000000?text=+) -> ![](https://placehold.it/15/037AA9/000000?text=+) - then we can see how it affects primary buttons (that would arguably be the point of biggest impact). @bia feel free to open a separate issue for that.

PR #169, opened 2018-04-11
Shifting turquoise to a darker AA compliant shade

bia:
Shifting  _turquoise_  a little darker to  _#037AA9_ because this colour passes WCAG AA both as a background with white text and as colored text on a  _whitesmoke_ background.

Completes #161.


bia:
Thanks for the review!

Alright, rebased the commits and turned down the darken effect.
Let's address the button color/behaviour in our future design system.

bia:
Yes, now that button backgrounds are generally becoming darker for AA compliance with white text, we might want to consider lightening on hover instead.
All noted in #174

PR #170, opened 2018-04-11
Remove metrics repeated across all series

### LINKED ISSUE #145: PrometheusGraph - remove metric labels that appear across all series
Related to https://github.com/weaveworks/service-ui/issues/2096#issuecomment-372344116.

fbarl:
Resolves #145.

### DIFF
--- a/src/components/PrometheusGraph/PrometheusGraph.js
+++ b/src/components/PrometheusGraph/PrometheusGraph.js
@@ -15,7 +15,9 @@ import {
   forEach,
   omit,
   size,
-  has,
+  isEmpty,
+  filter,
+  every,
   get,
 } from 'lodash';
 import { scaleLinear, scaleQuantize } from 'd3-scale';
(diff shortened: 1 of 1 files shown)


docs/js/Example.js
```diff
@@ -67,6 +67,17 @@ function getReadableValueType(type) {
   return type.name;
 }
 
+function getDefaultValue(defaultValue) {
+  if (!defaultValue) return '';
+
+  // If default value is a function, show only its name.
+  if (defaultValue.value.indexOf('function') === 0) {
```
foot: is this a js fn or a prom fn?

foot: and we can use includes! or is it contains....

fbarl: JS function :) I wanted to make the check a bit stricter and make sure the value begins with "function" instead of containing it at a random place :P

PR #185, opened 2018-04-19
Add prettier

jpellizzari:
Adds prettier and the changes as a result of running it. Really, its just me trying to keep my change count so it looks like I am more productive.


jpellizzari:
@fbarl I'm not able to reproduce. The range selector and prom graphs look the same as what is published: http://weaveworks-ui-components.s3-website-us-west-2.amazonaws.com/#/components/prometheusgraph

PR #187, opened 2018-04-23
Fix parentheses bugs

fbarl:
Fixes two bugs introduced in #185.

* Time Travel range selector
* Prometheus Graph time scale


foot:
Was it prettier do you think? It likes to fiddle w/ parens sometimes.

fbarl:
> Was it prettier do you think? It likes to fiddle w/ parens sometimes.

Not sure, @jpellizzari was playing with it :)

jpellizzari:
That one was me! eslint was complaining about the mixed operators. Nice catch @fbarl.

PR #188, opened 2018-04-23
[WIP] Remove SVG from timeline

### LINKED ISSUE #179: Move away from SVG rendering in timeline
Don't remember why SVG rendering was used for the timeline, probably there were some considerations of rendering more complex graph inside it initially, but everything we've planned adding in a foreseeable future seems easy to render with normal `div`s & `span`s.

The advantage of moving away from SVG would be probably faster rendering and not having to use `onResize` hooks anymore as we could use relative CSS positioning instead. Consequently, that would probably reduce flickering when switching between pages with Time Travel in them.

Related to #173.

fbarl:
Resolves #179.


fbarl:
Superseeded by #230.

PR #189, opened 2018-04-23
No longer have buttons all uppercase

### LINKED ISSUE #127: Button text should not be all caps
No reason it needs to be all caps

rndstr:
Closes #127

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -22,7 +22,6 @@ const StyledButton = styled('button')`
 
   /* Text */
   font-size: ${props => props.theme.fontSizes.normal};
-  text-transform: uppercase;
 
   /* Other */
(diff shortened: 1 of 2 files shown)


bowenli:
Do we want to change it to `capitalize` just to be safe? Will all the buttons currently be capitalized correctly if we just remove this?

rndstr:
> Do we want to change it to capitalize just to be safe?

For longer button texts we shouldn't use `capitalize` IMO. Also I would rather have it fixed in copy than dragging along an override. I'm currently glimpsing over all our buttons.

> Will all the buttons currently be capitalized correctly if we just remove this?

Guess that depends on the definition of "correct" :)

bowenli:
Ok that's fine. LGTM.

Issue #162, opened 2018-03-28
Organize z-index values semantically


fbarl:
Currently both `scope` and `service-ui` repos have a lot of pretty randomly assigned big `z-index` values and it makes it hard to keep track of what goes on top of what - see https://github.com/weaveworks/service-ui/pull/2217#issuecomment-376847947.

It might be useful to restrict the allowed values to a couple of categories defined in the theme - e.g. `notification`, `popup`, `tooltip`, `overlay` etc... and use only those. We could even try to enforce it via linting rules: https://www.npmjs.com/package/stylelint-no-z-index

PR #193, opened 2018-04-25
Collect z-index values semantically

fbarl:
First step in resolving #162.

### DIFF
--- a/src/theme/weave.js
+++ b/src/theme/weave.js
@@ -142,6 +142,16 @@ const weave = {
 
   borderRadius: '4px',
 
+  layer: {
+    front: 1,
+    toolbar: 2,
+    notification: 3,
+    alert: 4,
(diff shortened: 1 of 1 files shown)


src/theme/weave.js
```diff
@@ -142,6 +142,16 @@ const weave = {
...
 
+  layer: {
+    front: 1,
+    toolbar: 2,
+    notification: 3,
+    alert: 4,
+    dropdown: 5,
+    tooltip: 6,
+    modal: 7,
+  },
```
foot: Nice! Good thought, should modals beat notifications sometimes?  There're prolly not too many cases where this  really bad.

Dropdown over notification?...

Lets give it a go and then tweak it later.

foot: lgtm!

PR #196, opened 2018-04-28
Submit timestamp value only on enter or blur

rndstr:
Previously, when editing the timestamp value, the input field value was
replaced as soon as a loosely parsable date was found. This made editing
the timestamp cumbersome.

With this PR, the date will only be checked for validity and
subsequently sent to its parent only if Enter is pressed or the input loses
focus.


bowenli:
Fixes https://github.com/weaveworks/service-ui/issues/2237

Issue #202, opened 2018-05-07
Enforce theme font sizes


fbarl:
Although our `weave.js` theme includes some font sizes, they are not used that much across our code - I assume the two main reasons are:

a. There are only 3 sizes to choose from which still doesn't cover a lot of cases
b. Their usage is not enforced (or at least hinted by a warning)

```
fontSizes: {
  normal: '0.875em',
  large: '2em',
  xl: '2.827em',
}
```

I suggest the following:

1. Aggregate all the font sizes used across our app into `theme.fontSizes`
2. Force style linters to accept only variables passed to `font-size` property (like it's done for `z-index`)
3. Use only theme font sizes everywhere
4. _(not part of this issue)_ Start dropping some obsolete/ugly font sizes gradually while slowly converging to our unified design system, like we are currently doing with colors

cc @bia @miklosp @jpellizzari

jpellizzari:
I think we should go with a scale like this to cut down on the available options: `10, 12, 14, 18, 24, 36`. I think that gives us enough variation to get a nice hierarchy if we need it, without allowing for too much customization. The actual values aren't as important as limiting it to 6 or less.

Names (respectively): `tiny, small, normal, large, extra-large, heading`. I was thinking this would communicate some intent.

Ii also propose we build this into the `<Text />` component to make it very simple to follow the font-sizes.

bia:
The [typescale](http://type-scale.com/?size=14&scale=1.414&text=A%20Visual%20Type%20Scale&webfont=Proxima%20Nova&font-family=%27Proxima%20Nova%27%2C%20serif&font-weight=400&font-family-headers=&font-weight-headers=inherit&background-color=white&font-color=%23333) that we discussed on the Design Trello might still be a useful reference.

rndstr:
We should also use either `rem` or `px` IMHO. `em` is too much of a headache if people inadvertently nest font-sizes.

miklosp:
That's what I would recommend:
![image](https://user-images.githubusercontent.com/8489585/39861910-0c859f60-543a-11e8-918b-67a15018bb16.png)

miklosp:
@fbarl the grouping I would recommend is:
tiny, small > tiny
normal > small
large > normal
huge, enormous > large
gigantic > extra large
huge is used very sparingly (big numbers on instance homepage and dashboards)

fbarl:
[...]

PR #203, opened 2018-05-08
Add all theme font sizes

fbarl:
First step for #202.

image


fbarl:
> Not sure what a js.snap is, and why is the px value hardcoded there, but I assume that's the way it should be.

That's just a mockup file used only for testing and normally it's generated.

> I'm not sure whether using `px` instead of `rem` is a good practice, but that's another conversation.

I didn't have a strong preference there, went with `px` to avoid decimal values, but wouldn't mind changing it in subsequent PRs..

Issue #208, opened 2018-05-11
Enforce theme border radius


fbarl:
We have different values appearing across our code, including `2px`, `3px`, `4px`, `5px`, `8px`, `10px` and `50%`, but I think the only values we really need before enforcing the theme are `0`, `4px` (the current one) and `50%`, e.g.:

```
borderRadius.none: 0
borderRadius.soft: 4px
borderRadius.round: 50%
```

The cases like `border-radius: 0 4px 4px 0;` could then be split into two lines:

```
border-top-right-radius: ${props => props.theme.borderRadius.soft};
border-bottom-right-radius: ${props => props.theme.borderRadius.soft};
```

cc @bia @jpellizzari

PR #209, opened 2018-05-11
Use theme border radii

fbarl:
First step in resolving #208.

image

### DIFF
--- a/src/theme/weave.js
+++ b/src/theme/weave.js
@@ -144,7 +144,11 @@ const weave = {
     selected: `0px 0px 2px 2px ${colors.primary.lavender}`,
   },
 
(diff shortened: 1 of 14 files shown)


src/theme/weave.js
```diff
@@ -144,7 +144,11 @@ const weave = {
     selected: `0px 0px 2px 2px ${colors.primary.lavender}`,
   },
 
-  borderRadius: '4px',
+  borderRadius: {
+    none: '0',
+    soft: '4px',
+    round: '50%',
```
rndstr: what about naming this `circle` instead? that would not only describe to me clearer what the purpose of this radius is but also clarify what the target element is going to look like.

PR #222, opened 2018-05-17
PrometheusGraph legend items selection hook

### LINKED ISSUE #198: Move selection state out of PrometheusGraph component
We'd like to have legend selection & hover state shared across multiple charts, so we need to move the selection state up to a parent component.

We should provide a simple wrapped chart which preserves the current local selection state behaviour.

fbarl:
Resolves #198.

##### Changes

* Modified the way series keys are chosen so that they will differ from series names only if they repeat multiple times in the legend. So, when we know they don't repeat, we can assume `seriesKey === seriesName`.
* Added `onChangeLegendSelection` to `PrometheusGraph` to notify the outside world of legend selection changes.
* Added `selectedLegendKeys` prop to `PrometheusGraph` that overrides the local state whenever it's changed (this pattern has already been repeated in a lot of our React components that communicate their internal state with the outside world and it seems to work fine).
* Added an example of two correlated graphs which demonstrates how we could have a _global_ legend state selection system.

image

### DIFF
--- a/src/components/PrometheusGraph/PrometheusGraph.js
+++ b/src/components/PrometheusGraph/PrometheusGraph.js
@@ -21,6 +21,7 @@ import {
   get,
   reverse,
   indexOf,
+  noop,
 } from 'lodash';
 import { scaleLinear, scaleQuantize } from 'd3-scale';
 import { format, formatPrefix, precisionPrefix, precisionFixed } from 'd3-format';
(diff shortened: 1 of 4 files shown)


src/components/PrometheusGraph/PrometheusGraph.js
```diff
@@ -226,15 +228,21 @@ class PrometheusGraph extends React.PureComponent {
     if (this.props.multiSeries !== nextProps.multiSeries) {
       this.prepareMultiSeries(nextProps);
     }
+    if (this.props.selectedLegendKeys !== nextProps.selectedLegendKeys) {
```
leth: Do you need to do a shallow array comparison here?

fbarl: I think we may assume that whatever gets passed as `selectedLegendKeys` here comes originally from either a Redux or a local component state, both of which get their references updated as soon as their content changes.

It's an assumption we're making in (almost) all the places where we're checking if the props have changed really.

PR #224, opened 2018-05-17
Add TabSelect component

### LINKED ISSUE #190: Add TabSelector component
We currently use it in at least 2 different places:

* Workloads summary
image

* Deploy service locking reason
image

Related to https://github.com/weaveworks/service-ui/issues/2416.

jpellizzari:
Closes #190 

Adds a TabSelect component, borrowing heavily from the current implementation in service-ui.

I am kind of ignoring my own advice about introspecting into `children`, but I looked at some similar third party components and it seems like it might be okay. At least the `propType` validation will throw an error if any type of unreadable component shows up as a child.

API looks like this:

```javascript
      <TabSelect>
        <Tab label="A" name="a">
          Tab A
        </Tab>
        <Tab label="B" name="b">
          Tab B
        </Tab>
        <Tab label="C" name="c">
          Tab C
        </Tab>
      </TabSelect>
```

Renders:
screen shot 2018-05-17 at 3 44 26 pm

### DIFF
--- a/src/components/TabSelect/Tab.js
+++ b/src/components/TabSelect/Tab.js
@@ -1,0 +1,31 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+
+const Styled = component => styled(component)``;
+
+/**
+ * A tab component for use within a TabSelect component.
+ *  * See also:
+ * [TabSelect](/components/tabselect)
+ */
+
+class Tab extends React.PureComponent {
+  render() {
+    const { className, children } = this.props;
+    return <div className={className}>{children}</div>;
+  }
+}
+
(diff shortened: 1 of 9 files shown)


src/components/TabSelect/Tab.js
```diff
@@ -0,0 +1,25 @@
...
+class Tab extends React.Component {
+  render() {
+    const { className, children } = this.props;
+    return <div className={className}>{children}</div>;
+  }
+}
+
+Tab.propTypes = {
+  label: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
```
fbarl: Maybe add `className` here as well as comment descriptions?

PR #230, opened 2018-05-30
Move away from SVG rendering in Time Travel timeline

### LINKED ISSUE #179: Move away from SVG rendering in timeline

fbarl:
Resolves #179.

Nothing about the behaviour should have changed, only the underlying DOM tree elements.

### DIFF
--- a/src/components/TimeTravel/_TimelineRange.js
+++ b/src/components/TimeTravel/_TimelineRange.js
@@ -1,6 +1,17 @@
 import React from 'react';
 import moment from 'moment';
+import styled from 'styled-components';
 import PropTypes from 'prop-types';
+
+const TimelineRangeOverlay = styled.div.attrs({
+  style: ({ position }) => ({ transform: `translateX(${position}px)`})
+})`
+  background-color: ${props => props.color};
+  height: ${props => props.height}px;
+  width: ${props => props.width}px;
+  position: absolute;
+  opacity: 0.15;
+`;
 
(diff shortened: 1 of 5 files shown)


src/components/TimeTravel/_TimelineRange.js
```diff
@@ -1,7 +1,18 @@
 import React from 'react';
 import moment from 'moment';
+import styled from 'styled-components';
 import PropTypes from 'prop-types';
 
+const TimelineRangeOverlay = styled.div.attrs({
+  style: ({ position }) => ({ transform: `translateX(${position}px)`})
```
foot: `position` feels a bit ambiguous vs `x`. and I get `position` confused w/ the other css stuff.

and does this need to be a style attr?

fbarl: Makes sense!

> does this need to be a style attr?

Moving values that change a lot from component styles into `attr.style` is an optimization that _styled-components_ recommend, to prevent recreating the whole styled component object every time e.g. a position or size of an object changes.

PR #232, opened 2018-05-31
Add `width` prop to Dialog for content width

rndstr:


jpellizzari:
Can we accomplish this with styling instead? That element already has a `className="weave-dialog-window"`, so it can be styled externally.

rndstr:
@jpellizzari sure, I did use the class name in my case. I thought we might want to move away from using classNames in the future but CSS works for me as well.

jpellizzari:
>  thought we might want to move away from using classNames 

Still an open question for me. I can't decide if we should add classNames to all children in `styled-components`, or if we should NOT give the user that much control of styling. One could argue that changing the width of the `<Dialog />` is a style-guide "violation"...but maybe that is too restrictive.

rndstr:
> One could argue that changing the width of the <Dialog /> is a style-guide "violation"...but maybe that is too restrictive.

Could introduce size names instead such as `small` or `large`

PR #233, opened 2018-06-01
remove alert from DOM when not visible

### LINKED ISSUE #50: Alert: Collapse the <div> on close (or on visible="false")
At the moment setting `visible="false"` on an Alert just animates its opacity. Most often then not, I think we want the Alert div to go away instead of leaving a blank space on the page.

guyfedwards:
fixes #50


dlespiau:
It'd be interesting to know why you disagree with #50. To me `visible="false"` leaving the Alert size allocation visible on the page means the client code has to do something about it, making the component not entirely self-contained. Depends on how one sees the API contract of components I guess, imagine having several properties like that, we could en up with annoying logic in the client.

Another way to see it: are they cases where we want to keep the size of a `visible="false"` Alert in the page? I don't think so off hand, but maybe there are!

We don't necessarily need to remove the component from the DOM tree either, changing its size should be enough, it'll will be pruned early from the rendering tree(s).

jpellizzari:
There are already patterns for conditionally removing an item from the DOM, no need to build that into our components' API. Also, the animation was a deliberate choice to avoid being abrupt about showing/hiding the alert.

> Another way to see it: are they cases where we want to keep the size of a visible="false" Alert in the page?

This is a good way to prevent the other content from sliding up and down when the alert is shown/not shown. You can also just use `absolute` positioning to avoid taking up space on the page.

> We don't necessarily need to remove the component from the DOM tree either, changing its size should be enough, it'll will be pruned early from the rendering tree(s).

Not sure I follow here.

dlespiau:
I definitely don't know the react patterns, so yes, it may be very idiomatic to leave the element in the page :)

For the size bit, I was trying to say that reducing the element size to 0 is a way to not render it while keeping it in the DOM tree, the browser rendering code has optimisations to remove non visible elements very early.

guyfedwards:
> The consumer of the <Alert /> always has the option to remove it from the DOM themselves. 
Valid point, think this answers the questions around remove or not.
I'm still not up to speed with all the places alerts are used so didn't see all the use-cases. Will close this and proceed with the title/icon additions and leave visibility as is

PR #252, opened 2018-06-11
add secondary styling for TabSelect

### LINKED ISSUE #251: Create secondary style TabSelect
Add new style of TabSelec with transparent background and no border around the content. 
This should be added as a theme/style option for TabSelect as it is likely to be reused.

guyfedwards:
Currently uses `theme.colors.sand` which is the background used across the site. Unsure if there are needs for different backgrounds in which case we can make this configurable?

image

Fixes #251

### DIFF
--- a/src/components/TabSelect/TabSelect.js
+++ b/src/components/TabSelect/TabSelect.js
@@ -8,16 +8,23 @@ import TabButton from './_TabButton';
 
 const TabButtons = styled.div``;
 
-const TabContent = styled.div`
-  padding: 20px;
-  background-color: ${props => props.theme.colors.white};
-
-  ${props => `
+const borders = props => `
     border-top-right-radius: ${props.theme.borderRadius.soft};
     border-bottom-right-radius: ${props.theme.borderRadius.soft};
     border-bottom-left-radius: ${props.theme.borderRadius.soft};
     border: 1px solid ${props.theme.colors.athens};
-  `};
+`;
+
+const bordersOnlyTop = props => `
+  border-top: 1px solid ${props.theme.colors.athens}
+`;
+
+const TabContent = styled.div`
+  padding: 20px;
+  background-color: ${props =>
+    props.secondary ? props.theme.colors.sand : props.theme.colors.white};
+
+  ${props => (props.secondary ? bordersOnlyTop(props) : borders(props))};
 `;
 
 const Styled = component => styled(component)``;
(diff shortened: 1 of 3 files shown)


src/components/TabSelect/_TabButton.js
```diff
@@ -3,6 +3,14 @@ import styled from 'styled-components';
 
 const TabName = styled.span``;
 
+const getBgColor = (selected, secondary, theme) => {
+  if (secondary) {
+    return theme.colors.sand;
```
fbarl: Any reason why we'd prefer a fixed color here instead of `'transparent'`?

src/components/TabSelect/TabSelect.js
```diff
@@ -8,16 +8,23 @@ import TabButton from './_TabButton';
...
+`;
+
+const bordersOnlyTop = props => `
+  border-top: 1px solid ${props.theme.colors.athens}
+`;
+
+const TabContent = styled.div`
+  padding: 20px;
+  background-color: ${props =>
+    props.secondary ? props.theme.colors.sand : props.theme.colors.white};
```
fbarl: Also here - why not use `transparent` instead of `colors.sand`?

src/components/TabSelect/example.js
```diff
@@ -6,18 +6,33 @@ import Tab from './Tab';
...
+          </Tab>
+          <Tab label="B" name="b">
+            Tab B
+          </Tab>
+          <Tab label="C" name="c">
+            Tab C
+          </Tab>
+        </TabSelect>
+      </Example>
+      <Example>
```
fbarl: Now that we have two examples here, maybe you can add `<Info>` titles above each example as in e.g. Time Travel:

image

src/components/TabSelect/_TabButton.js
guyfedwards: Great point and kinda what I was alluding to in the description regarding background color. We have to set the color explicitly for the tab button itself otherwise the top border of the content shows.

src/components/TabSelect/TabSelect.js
guyfedwards: same as above

src/components/TabSelect/example.js
guyfedwards: 👌

src/components/TabSelect/_TabButton.js
guyfedwards: We could add a prop for `secondaryBg` if there is a need to override `theme.colors.sand` but I can't see any way to use transparent unfortunately

fbarl: Ok I see what you mean now and I agree we should only try to address this if the use case occurs. :)

PR #257, opened 2018-06-12
add Code component for showing/copying snippets

### LINKED ISSUE #244: Add CodeBlock component
Used in a bunch of places in Weave Cloud.

Should support both shell commands and raw file format, with standardized coloring.

image

guyfedwards:
I have tried to simplify the API and usage to be more intuitive although for more complex styled/dynamic setups there is still some jsx wrangling required.

Couple of things to discuss:
 - current implementation has the content scroll when it overflows. I think this is the sanest option, but visually will change across different OS/Browsers?
 - Do we want to shrink the copy button so it occupies less of the code space?

image

image

image

fixes #244

### DIFF
--- a/src/components/Code/Code.js
+++ b/src/components/Code/Code.js
@@ -1,0 +1,150 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { isFunction, isString, trim, noop } from 'lodash';
+
+const CopyNotice = styled.div`
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: 10px 15px;
+  border-radius: ${props => props.theme.borderRadius.soft};
+  background-color: ${props => props.theme.colors.purple800};
+  transition: opacity 300ms ease;
+  opacity: 0;
+  font-size: ${props => props.theme.fontSizes.small};
+  color: ${props => props.theme.colors.purple50};
+`;
+
+const CodeWrapper = styled.div`
+  position: relative;
+  padding: 10px 15px;
+  background-color: ${props => props.theme.colors.purple800};
+  box-sizing: border-box;
+  border-radius: ${props => props.theme.borderRadius.soft};
+  cursor: pointer;
+
+  &:hover ${CopyNotice} {
+    opacity: 0.9;
+  }
+`;
+
+const Content = styled.div`
+  flex-grow: 1;
+  overflow: auto;
+  color: ${props => props.theme.colors.purple50};
+`;
+
+const Pre = styled.pre`
+  margin: 10px 0;
+  font-family: ${props => props.theme.fontFamilies.monospace};
+  font-size: ${props => props.theme.fontSizes.small};
+`;
+
+const trimString = node => (isString(node) ? trim(node) : node);
+
+/**
+ * Code allows for easy rendering of code snippets which can easliy be copied to
+ * the clipboard by clicking the element or selecting a portion of the code
+ */
+class Code extends Component {
+  static displayName = 'Code';
+
+  state = {
+    isCopying: false,
+  };
+
+  componentDidUpdate = () => {
+    // reselect text after re-render
+    if (this.state.selectedRange) {
+      const prevRange = this.state.selectedRange;
+      const sel = window.getSelection();
+      sel.removeAllRanges();
+      sel.addRange(prevRange);
+      // clear it for the next interaction
+      this.setState({ selectedRange: null });
+    }
+  };
+
+  handleCopyClick = () => {
+    // Create a textarea element in which the text of the code will be copied.
+    // Set the value of the textarea, select it to set it as the active element,
+    // then copy the value of that element to the clipboard.
+    // It's 2017 and this is still the best way to copy text on button click...
+    if (isFunction(this.props.onCopy)) {
+      this.props.onCopy();
+    }
+    const selection = document.getSelection();
+    const selectionString = selection.toString();
+    // if user has selected text, save that to state so it can be reselected
+    if (selectionString !== '') {
+      this.setState({
+        selectedRange: selection.getRangeAt(0),
+      });
+    }
+    // console.dir(document.getSelection().anchorNode);
+    const code =
+      selectionString === '' ? this.preNode.textContent : selectionString;
+    const txtArea = document.createElement('textarea');
+    // Safari doesn't allow for assigning an object literal to `style`.
+    txtArea.class = 'hidden-textarea';
+    // Make sure the code ends in a newline to execute all the commands in multiline code.
+    if (code.charAt(code.length - 1) === '\n') {
+      txtArea.value = code;
+    } else {
+      txtArea.value = `${code}\n`;
+    }
+    document.body.appendChild(txtArea);
+
+    txtArea.select();
+
+    try {
+      document.execCommand('copy');
+    } catch (e) {
+      throw e;
+    }
+    document.body.removeChild(txtArea);
+
+    // show the Copied to clipboard notice temporarily
+    this.setState({ isCopying: true });
+    setTimeout(() => {
+      this.setState({ isCopying: false });
+    }, 3000);
+  };
+
+  render() {
+    const { children } = this.props;
+    const { isCopying } = this.state;
+
+    const copyText = isCopying ? 'Copied to clipboard' : 'Copy to clipboard';
+    const copyIcon = isCopying ? 'files-o' : 'check';
+
+    return (
+      <CodeWrapper onClick={this.handleCopyClick}>
+        <Content>
+          <Pre innerRef={e => (this.preNode = e)}>
+            {isFunction(children) ? children() : trimString(children)}
+          </Pre>
+        </Content>
+
+        <CopyNotice>
+          <i className={`fa fa-${copyIcon}`} /> {copyText}
+        </CopyNotice>
+      </CodeWrapper>
+    );
+  }
+}
+
+Code.propTypes = {
+  // Children can be anything that React can render
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+
+  // onCopy will be called when the CodeWrapper is clicked
+  onCopy: PropTypes.func,
+};
+
+Code.defaultProps = {
+  onCopy: noop,
+};
+
+export default Code;
--- a/src/components/Code/Code.test.js
+++ b/src/components/Code/Code.test.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import enzyme, { shallow, mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { withTheme } from '../../utils/theme';
+import Code from '.';
+
+enzyme.configure({ adapter: new Adapter() });
+
+document.execCommand = () => {};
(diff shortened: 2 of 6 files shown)


fbarl:
> Current implementation has the content scroll when it overflows. I think this is the sanest option, but visually will change across different OS/Browsers?

I think this is good enough for now.

> Do we want to shrink the copy button so it occupies less of the code space?

I just talked about it with @bia and we both think it might be better if we just got rid of that button :) What we could have instead is some sort of a popup button on text selection - I'll let her post some screenshots.

That way we would solve both the vertical spacing and potentially partial selection issue (see https://github.com/weaveworks/service-ui/issues/2696).

Nice examples btw! Will take a look at the code once we settle on the _Copy_ button.

guyfedwards:
👍 Cool beans. Agree about the button, I had been wondering about something like this:
normal:
image
hover:
image

bia:
Sounds good.
I was talking to @fbarl about even showing no overlay at all, then on mouse down the word 'Copied' or 'Copied to clipboard' in the top right corner. Then the person could continue to select normally within the code block if they needed one piece in particular.

What do you think? Shall I come up with a mockup tomorrow?

bia:
How does this feel?

copycode
copycode2

I would also keep the slightly rounded corners - it helps a block this dark not to seem so visually heavy.

guyfedwards:
@bia Rounded corners 👍 forgot to add them back, couple of questions: 
- Is that triggered on clicking the code block itself? (are we implementing the partial selection as per https://github.com/weaveworks/service-ui/issues/2696)
- Is it obvious to the user that thats how it works 🤔 ?

bia:
>Is that triggered on clicking the code block itself? (are we implementing the partial selection as per weaveworks/service-ui#2696)

We are going to bring back partial selection.

>Is it obvious to the user that thats how it works 🤔, some indication that they can click it, an icon maybe?

Good point.
What if we display this as a hover state
copycode3

and this on mouse down.
copycode2


You could continue to drag and partially select to manually copy a portion, but at this point all of the text is in your clipboard.

guyfedwards:
You can now click anywhere on the code block to copy all the text, or select some text and only that text will be selected.

src/components/Code/example.js
```diff
@@ -0,0 +1,141 @@
+import React from 'react';
+import { trim } from 'lodash';
+import styled from 'styled-components';
+
+import { Example, Info } from '../../utils/example';
+import { Input } from '..';
+import Code from '.';
+
+const Highlight = styled.span`
+  color: ${props => props.theme.colors.accent.blue};
```
fbarl: Please rebase this PR to capture the latest color naming changes.. Sorry for that :/

src/components/Code/example.js
```diff
@@ -0,0 +1,141 @@
...
+            kubectl create secret generic -n weave cloudwatch
+            --from-literal=access-key-id=xyz
+            --from-literal=secret-access-key=abc
+          </Code>
+        </Example>
+
+        <Info>Multiline file</Info>
+        <Example>
+          <Code isFile onCopy={this.onCopy}>
+            {trim(`
```
fbarl: Perhaps it would make sense to wrap `children` in `trim` internally if the string is passed as the only param, i.e. `isString(children)`.

src/components/Code/example.js
```diff
@@ -0,0 +1,141 @@
...
+  --from-literal=access-key-id=xyz \\
+  --from-literal=secret-access-key=abc
+          `)}
+          </Code>
+        </Example>
+
+        <Info>Dynamic content in template string</Info>
+        <Example>
+          <Code isFile onCopy={this.onCopy}>
+            {() =>
```
fbarl: Is there a use case for passing a function? In a sense that if it wasn't a function, something would not work (or be much more complicated)?

src/components/Code/example.js
```diff
@@ -0,0 +1,141 @@
...
+                {'}'}
+              </div>
+              <div>{'}'}</div>
+            </div>
+          </Code>
+        </Example>
+      </div>
+    );
+  }
+}
```
fbarl: Great examples! :+1:

src/components/Code/Code.test.js
```diff
@@ -0,0 +1,41 @@
...
+import Adapter from 'enzyme-adapter-react-16';
+import { withTheme } from '../../utils/theme';
+import Code from '.';
+
+enzyme.configure({ adapter: new Adapter() });
+
+document.execCommand = () => {};
+document.getSelection = () => ({
+  toString: () => '',
+});
```
fbarl: Do these get properly rolled back after the tests are finished?

src/components/Code/Code.js
```diff
@@ -0,0 +1,107 @@
...
+
+const CopyNotice = styled.div`
+  position: absolute;
+  top: 5px;
+  right: 5px;
+  transition: opacity 300ms ease;
+  opacity: 0;
+  font-size: ${props => props.theme.fontSizes.tiny};
+  color: ${props => props.theme.colors.athensGray};
+`;
```
fbarl: I suggest a couple of styling changes here that IMO make it look better:

* Set both `top` and `right` to `0`
* Add padding `10px 15px` instead, like in the wrapper element
* Add the same `borderRadius` and `backgroundColor` like the wrapper element as well
* Set the full opacity to `0.9` instead of `1` below, so that the text overlay over the code looks a bit better
* Set font size to `small` instead of `tiny` after all

Basically, I think the text shouldn't be that much stuck to the corner and increasing the padding will potentially cause overlap with the text (which should be fine), therefore the rest of the changes.

cc @bia

src/components/Code/Code.js
```diff
@@ -0,0 +1,107 @@
...
+
+const Pre = styled.pre`
+  margin: 10px 0;
+  font-family: ${props => props.theme.fontFamilies.monospace};
+  font-size: ${props => props.theme.fontSizes.small};
+`;
+
+class Code extends Component {
+  static displayName = 'Code';
```
fbarl: Wouldn't that already be a default?

src/components/Code/Code.js
```diff
@@ -0,0 +1,107 @@
...
+    } catch (e) {
+      throw e;
+    }
+    document.body.removeChild(txtArea);
+
+    // show the Copied to clipboard notice temporarily
+    this.setState({ isCopying: true });
+    setTimeout(() => {
+      this.setState({ isCopying: false });
+    }, 1000);
```
fbarl: I think I'd prefer this interval to be a bit longer - perhaps 3 seconds?

src/components/Code/Code.js
```diff
@@ -0,0 +1,107 @@
...
+
+        <CopyNotice>
+          <i className={`fa fa-${copyIcon}`} /> {copyText}
+        </CopyNotice>
+      </Root>
+    );
+  }
+}
+
+export default Code;
```
fbarl: `propTypes` (and `defaultProps`) are missing here - we've decided to always add those in this repo, please see any other component around for how to do it.

src/components/Code/Code.js
```diff
@@ -0,0 +1,107 @@
...
+  color: ${props => props.theme.colors.athensGray};
+`;
+
+const Pre = styled.pre`
+  margin: 10px 0;
+  font-family: ${props => props.theme.fontFamilies.monospace};
+  font-size: ${props => props.theme.fontSizes.small};
+`;
+
+class Code extends Component {
```
fbarl: Component description (a comment above this line) that shows up in the _Examples_ page would also be nice.

src/components/Code/Code.test.js
```diff
@@ -0,0 +1,41 @@
...
+    );
+    expect(onCopy).not.toHaveBeenCalled();
+    wrapper.find('Code__CopyNotice').simulate('click');
+    expect(onCopy).toHaveBeenCalledTimes(1);
+  });
+
+  it('should render correctly', () => {
+    expect(wrapper.dive()).toMatchSnapshot();
+  });
+});
```
fbarl: Thanks for creating the tests! :ok_hand:

src/components/Code/Code.js
```diff
@@ -0,0 +1,107 @@
...
+  position: absolute;
+  top: 5px;
+  right: 5px;
+  transition: opacity 300ms ease;
+  opacity: 0;
+  font-size: ${props => props.theme.fontSizes.tiny};
+  color: ${props => props.theme.colors.athensGray};
+`;
+
+const Root = styled.div`
```
fbarl: `Root` is an odd name for an element (it gets shown in `className`) - normally we'd name this one `CodeWrapper` or something similar :)

src/components/Code/example.js
guyfedwards: Initially I had this but might there be an occasion where you want space before/after?

src/components/Code/example.js
guyfedwards: Allows for using template string formatting with dynamic values as opposed to having to deal with indentation in JSX

src/components/Code/Code.test.js
guyfedwards: Great question, I believe they do as each test is isolated.
They are purely there to stop Jest barfing about `document.execCommand` not being a function as it is neither are mocked in JSDom.

guyfedwards: Just checked and the tests are isolated so this will not affect other tests regardless

src/components/Code/example.js
fbarl: I don't think so - if we currently don't have such a use case anywhere in the Service UI, I think you can assume it's not likely to happen :)

src/components/Code/example.js
fbarl: Ok I see, was wondering about the particular use case in Service UI, but found `FluxSetupKubectlCommand` :)

src/components/Code/Code.js
```diff
@@ -0,0 +1,150 @@
...
+Code.propTypes = {
+  // Children can be anything that React can render
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+
+  // onCopy will be called when the CodeWrapper is clicked
+  onCopy: PropTypes.func,
+};
+
+Code.defaultProps = {
+  onCopy: () => {},
```
fbarl: There is `lodash.noop` which is maybe a tiny bit cleaner :)

src/components/Code/Code.js
```diff
@@ -0,0 +1,150 @@
...
+          <i className={`fa fa-${copyIcon}`} /> {copyText}
+        </CopyNotice>
+      </CodeWrapper>
+    );
+  }
+}
+
+Code.propTypes = {
+  // Children can be anything that React can render
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
```
fbarl: Hmm, so even strings are seen as `PropTypes.node`? Seems so, I don't see React complaining in the examples :)

guyfedwards: Yea node is anything that can be rendered, however functions aren't technically renderable until they are invoked so requires a separate type :man_shrugging:

src/components/Code/Code.js
guyfedwards: certainly is :+1:

PR #271, opened 2018-06-19
Pass any props down to html button

foot:
- e.g. title, data-* etc

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import { omit } from 'lodash';
 
 import { fromAtoms } from '../../utils/theme';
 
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.js
```diff
@@ -53,20 +54,19 @@ class Button extends React.PureComponent {
...
-      disabled,
       danger,
       selected,
-      style,
-      className,
+      disabled,
+      ...otherProps,
     } = this.props;
 
+    const buttonProps = omit(otherProps, ['onClick']);
```
fbarl: Could we also omit the `style` prop? :) It's not cool that it's overridable (and also couldn't find any place in the code where it's actually being overriden).

foot: > couldn't find any place in the code where it's actually being overriden

we have the best coders!

I agree we shouldn't use it... unless its an emergency ofc. I guess because we accept className you can styled(Button) it...

PR #280, opened 2018-06-22
add multiCommand prop to Code

### LINKED ISSUE #278: Code component should have better formatting for multi-command snippets
Currently multi commands look the same as a single command split across multiple lines. We should add some vertical spacing between lines.

guyfedwards:
visual differentiation between single command split across multiple
lines and multiple commands to be run in sequence

fixes #278 
image

Have added three new examples for the multiCommand variations, not sure if they are really needed, would probably be fine with one additional example? @fbarl thoughts?

### DIFF
--- a/src/components/Code/Code.js
+++ b/src/components/Code/Code.js
@@ -65,7 +65,56 @@ const Pre = styled.pre`
   font-size: ${props => props.theme.fontSizes.small};
 `;
 
+const PaddedLine = styled.div`
+  &:not(:last-child) {
+    padding-bottom: 8px;
+  }
+
+  /* required to make jsx children work with adding '\n' in multiLine */
+  & > div {
+    display: inline;
+  }
+`;
+
 const trimString = node => (isString(node) ? trim(node) : node);
+
+const formatSingleCommand = children =>
+  isFunction(children) ? children() : trimString(children);
+
+function formatMultiString(string) {
+  return trim(string)
+    .split('\n')
+    .map((line, i) => (
+      <PaddedLine key={i}>
+        {line}
+        {'\n'}
+      </PaddedLine>
+    ));
+}
+
+function formatMultiCommand(raw) {
+  let children = raw;
+
+  if (isFunction(children)) {
+    children = children();
+  }
+  const count = React.Children.count(children);
+
+  if (count === 1 && isString(children)) {
+    return formatMultiString(children);
+  }
+
+  if (count === 1) {
+    return children;
+  }
+
+  return React.Children.map(children, (child, i) => (
+    <PaddedLine key={i}>
+      {child}
+      {'\n'}
+    </PaddedLine>
+  ));
+}
 
 /**
  * Code allows for easy rendering of code snippets which can easliy be copied to
@@ -151,7 +200,7 @@ class Code extends Component {
   };
 
   render() {
-    const { children } = this.props;
+    const { children, multiCommand } = this.props;
     const { isCopying, isHovered } = this.state;
 
     const copy =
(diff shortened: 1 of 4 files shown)


src/components/Code/Code.js
```diff
@@ -41,8 +41,57 @@ const Pre = styled.pre`
   font-size: ${props => props.theme.fontSizes.small};
 `;
 
+const PaddedLine = styled.div`
+  &:not(:last-child) {
+    padding-bottom: 8px;
+  }
+
+  /* required to make jsx children work with adding '\n' in multiLine */
+  & > div {
```
fbarl: This assumes something about the DOM structure of children passed to `<Code />` component, so I'd rather avoid it if possible. Would a `& > *` selector do the trick?

src/components/Code/Code.js
```diff
@@ -41,8 +41,57 @@ const Pre = styled.pre`
...
 const trimString = node => (isString(node) ? trim(node) : node);
 
+const formatSingleCommand = children =>
+  isFunction(children) ? children() : trimString(children);
+
+function formatMultiString(string) {
+  return trim(string)
+    .split('\n')
+    .map((line, i) => (
+      <PaddedLine key={i}>
```
fbarl: Passing indices as keys is not recommendable - how about `key={line}`?

src/components/Code/Code.js
fbarl: Actually wouldn't this already work if just `{'\n'}` was put outside of `<PaddedLine>` block below?

src/components/Code/Code.js
```diff
@@ -41,8 +41,57 @@ const Pre = styled.pre`
...
+  }
+  const count = React.Children.count(children);
+
+  if (count === 1 && isString(children)) {
+    return formatMultiString(children);
+  }
+
+  if (count === 1) {
+    return children;
+  }
```
fbarl: Do we even need this case for what is supposed to be a `multiCommand`? If yes, maybe put it inside one `if (count === 1)` block?

guyfedwards: Fair point, more just a guard against having the prop on with a single child. Moved into single `if` block

src/components/Code/Code.js
guyfedwards: Would require another wrapper for adjacent jsx elements as we don't have `React.Fragment` yet. Have changed to `*` selector

fbarl: Ok - I think having another wrapper wouldn't be such a big deal either (in case this proves to be non-ideal) :)

src/components/Code/Code.js
guyfedwards: Can do, not sure if there are any implications of the key being an arbitrary string length/content wise? 
The other key in `formatMultiCommand` is also index but we don't have a unique identifier such as line, not sure what the usual form is in such a situation? 
While not ideal, using index only really becomes as issue when performance is mission critical which I wouldn't say it was in this component.

guyfedwards: > Can do, not sure if there are any implications of the key being an arbitrary string length/content wise?
seems to be fine and is escaped by react so no worries about user input 🎉

src/components/Code/Code.js
```diff
@@ -84,8 +84,8 @@ const formatSingleCommand = children =>
 function formatMultiString(string) {
   return trim(string)
     .split('\n')
-    .map((line, i) => (
-      <PaddedLine key={i}>
+    .map(line => (
+      <PaddedLine key={`"${line}`}>
```
fbarl: Why the `"` and not simply `key={line}`?

PR #283, opened 2018-06-25
animate copy text to check icon Code component

guyfedwards:
code_anim_2

### DIFF
--- a/src/components/Code/Code.js
+++ b/src/components/Code/Code.js
@@ -1,7 +1,20 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
+import styled, { keyframes } from 'styled-components';
 import { isFunction, isString, trim, noop } from 'lodash';
+
+const scale = keyframes`
+  0% {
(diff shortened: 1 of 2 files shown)


src/components/Code/Code.js
```diff
@@ -1,8 +1,21 @@
...
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
+import styled, { keyframes } from 'styled-components';
 import { isFunction, isString, trim, noop } from 'lodash';
 
+const scale = keyframes`
+  0% {
+    transform: scale(0)
+  }
+
```
fbarl: OCD nit - get rid of this empty line? same below (line `35`) :P

fbarl:
> If I click on a code block, then click on another one, then quickly go back to hover on the first code block, the checkmarks is still being displayed.

Noticed that one as well, it's due to `3s` timeout and keeping the state there, don't feel strongly about it  :)

> The checkmark should only be displayed for a block of text which is currently in my clipboard.

The text still stays in your clipboard though :P

PR #287, opened 2018-06-26
arc animations when mounting/unmounting

guyfedwards:
animation wouldnt happen if props used directly so set them in state

### DIFF
--- a/src/components/ResourceDial/ResourceDial.js
+++ b/src/components/ResourceDial/ResourceDial.js
@@ -104,8 +104,24 @@ const dialSpring = value =>
   spring(value, { stiffness: 50, damping: 13, precision: 0.01 });
 
 class ResourceDial extends React.PureComponent {
+  constructor(props, context) {
+    super(props, context);
+
+    this.state = {
+      value: null,
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // setState is async so triggers another render allowing animation to happen
+    this.setState({
+      value: nextProps.value,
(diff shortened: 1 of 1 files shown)


src/components/ResourceDial/ResourceDial.js
```diff
@@ -104,8 +104,27 @@ const dialSpring = value =>
...
+
+    this.state = {
+      value: null,
+    };
+  }
+
+  // set value in state instead of directly using props to allow animation
+  // even when showing/hiding the component
+  componentWillReceiveProps(nextProps) {
+    if (isEqual(nextProps.value, this.props.value)) {
```
fbarl: `isEqual` -> `!isEqual`? :)

I don't think you need to use `isEqual` anyway, `nextProps.value !== this.props.value` would be just fine in this case.

Actually I think this piece of logic is corrupt because if you put `!==` there, it will never set the state unless the value changes :/ Even if the check is not there at all, `componentWillReceiveProps` will never be called if none of the props ever change.

What we probably need is something like this:

```
componentDidMount() {
  // Set value in state to trigger the animation only in the second render loop
  setTimeout(() => {
    this.setState({ value: this.props.value });
  }, 0);
}

componentWillReceiveProps(nextProps) {
  this.setState({ value: nextProps.value });
}
```

guyfedwards: I think the heat is making me see and write things that are just wrong 😄 🤦‍♂️

PR #290, opened 2018-06-27
add small option for TabSelect

guyfedwards:
used for code snippets in k8s onboarding, used for more inline tabs within
other page content

image

### DIFF
--- a/src/components/TabSelect/_TabButton.js
+++ b/src/components/TabSelect/_TabButton.js
@@ -15,8 +15,9 @@ const Styled = component => styled(component)`
   cursor: pointer;
   margin-bottom: -1px;
   margin-right: 5px;
-  padding: 10px 20px;
-  font-size: ${props => props.theme.fontSizes.large};
+  padding: ${props => (props.small ? '5px 10px' : '10px 20px')};
+  font-size: ${props =>
+    props.small ? props.theme.fontSizes.normal : props.theme.fontSizes.large};
   outline: 0;
 
   ${({ selected, secondary, theme }) => `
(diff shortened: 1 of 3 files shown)


src/components/TabSelect/_TabButton.js
```diff
@@ -15,8 +15,8 @@ const Styled = component => styled(component)`
   cursor: pointer;
   margin-bottom: -1px;
   margin-right: 5px;
-  padding: 10px 20px;
-  font-size: ${props => props.theme.fontSizes.large};
+  padding: ${props => (props.small ? '5px 10px' : '10px 20px')};
+  font-size: ${props => props.theme.fontSizes[props.small ? 'small' : 'large']};
```
fbarl: I think we should use `normal` font size with `props.small`, also I'd rather do something like

```
font-size: ${props => props.small ? props.theme.fontSizes.normal : props.theme.fontSizes.large};
```

so that those are easier to search and detect, e.g. see https://github.com/weaveworks/service-ui/pull/2803.

src/components/TabSelect/_TabButton.js
```diff
@@ -15,8 +15,8 @@ const Styled = component => styled(component)`
   cursor: pointer;
   margin-bottom: -1px;
   margin-right: 5px;
-  padding: 10px 20px;
-  font-size: ${props => props.theme.fontSizes.large};
+  padding: ${props => (props.small ? '5px 10px' : '10px 20px')};
```
fbarl: Right now the content is somewhat missaligned with the tab buttons with `props.small`:

image

Would be good to use the same content paddings to get something like:

image

PR #295, opened 2018-07-11
Data table improvements

jpellizzari:
Adds some extra functionality around sorting and fixes incorrect carets.

### DIFF
--- a/src/components/DataTable/_Header.js
+++ b/src/components/DataTable/_Header.js
@@ -9,8 +9,7 @@ const StyledHeader = component => styled(component)`
   width: ${props => props.width};
 
   &:hover {
-    color: ${props =>
-      props.sortable ? props.theme.colors.blue700 : 'inherit'};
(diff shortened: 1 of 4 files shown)


src/components/DataTable/_Header.js
```diff
@@ -30,13 +29,7 @@ class Header extends React.Component {
...
         {children}{' '}
         {order &&
-          sortable && (
-            <i
-              className={
-                order === 'asc' ? 'fa fa-caret-up' : 'fa fa-caret-down'
-              }
-            />
-          )}
+          sortable && <i className={order === 'desc' ? 'fa fa-caret-up' : 'fa fa-caret-down'} />}
```
rndstr: There doesn't seem a common consensus online which way is better. What's _your_ reasoning for seeing the previous version as incorrect?

Both ways have their merits I guess

- `asc ▲` small end vs big end
- `asc ▼` is pointing in the direction to read the list

since `asc` is usually the preferred way to read a list, having a triangle pointing "backwards" for ascending conveys to me that there might be something off while it isn't.

Issue #178, opened 2018-04-18
Revise status colors


bia:
Now that we've pruned a lot of our palette, our status colors are need of a little love
* `#9a9a9a` _disabled_ 
* `#d4ab27` _warning_ 
* `#38bd93` _success_ 
* `#049cd7` _info_ 
* `#c7254e` _error_ 

@fbarl and I have discussed merging `#037aa9` _turquoise_  into `#049cd7` _info_ because _turquoise_ is AA compliant both as a foreground and background

In accordance with #171, we might switch`#c7254e` _error_ to something like `#AD3131` _wellRead_ which is less purple shifted and AA compliant both as a foreground and background, like _turquoise_.

@LiliC and I have discussed having two different levels of `#d4ab27` _warning_, one lighter yellow and one more orange so that we can avoid using red unless a problem is very serious.

We did voice the concern that these changes might make the status colors no longer seem like a palette group, but I'd like to address that if and when it happens.

fbarl:
This issue has almost been resolved now - see https://github.com/weaveworks/ui-components/issues/148#issuecomment-413875368.

fbarl:
This issue has been fully resolved.

PR #298, opened 2018-07-16
Get rid of status.error color in favor of different shades of theme orange

### LINKED ISSUE #268: Use a shade of accent.orange for danger buttons instead of status.error
Accent colors for danger buttons would be more appropriate - also might be good to differentiate between errors and dangerous actions colorwise.

fbarl:
Resolves #268.

Affects:

* Danger button
* Error alert
* All error messages

##### Before

image
image

##### After

image
image

cc @bia


bia:
Resolves #178 (at least partially?)

fbarl:
@bia The contrast between `orange600` and `white` is actually AA compliant already :)

Issue #255, opened 2018-06-12
Export linting rules


fbarl:
Currently we're using separate (although similar) linting rules between our UI repos. It would be good to have one source of truth for all the repos: `ui-components`, `scope`, `service-ui`, for obvious styling reasons as well as for `service-ui` proxying to work smoothly - see https://github.com/weaveworks/service-ui/issues/2495#issue-321155370.

@foot and I talked about it and we thought it would be good to start with `ui-components` repo exporting its rules and then using them in `scope` and `service-ui`. Later on, we might want to move the rules into a separate repo.

The files that we'd need to export:

* `.eslintrc`
* `.stylelintrc`
* `.sass-lint.yml`?

guyfedwards:
Previously I have had success with creating a custom preset `eslint-preset-weaveworks` wrapping all this up. I also creates a forum for people to propose changes to the linting rather than additions being lost within other PRs.

Same for stylelint, we might be able to bundle `.sass-lint.yml` in there as well so that all style linting is in one package.

PR #300, opened 2018-07-16
align eslint versions across projects & fix lints

guyfedwards:
For reviewing purposes the bits to check are, `package.json`, `.eslintrc` and some `defaultProps` declarations. Most of everything else was basic lint fixing with `--fix`, `yarn prettify` and manual fixes for some of the unfixable stuff.

- align rulesets
- add prettier plugin to config
- only prettify src/styleguide/docs

related #255

### DIFF
--- a/src/components/TabSelect/TabSelect.js
+++ b/src/components/TabSelect/TabSelect.js
@@ -104,8 +104,15 @@ TabSelect.propTypes = {
   /**
    * Children of `TabSelect` must be a `Tab` component
    */
+  // eslint can't figure out that this is required because of custom checker
+  // function and so requires a defaultProp be set
+  // eslint-disable-next-line react/require-default-props
   children: (props, propName) => {
     let error = null;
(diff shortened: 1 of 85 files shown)


src/components/TabSelect/TabSelect.js
```diff
@@ -104,8 +104,13 @@ TabSelect.propTypes = {
   /**
    * Children of `TabSelect` must be a `Tab` component
    */
+  // eslint-disable-next-line
```
fbarl: Why? I think it's always good to go with `// eslint-disable name-of-specific-rule` to make that more clear..

fbarl:
So I assume the next step is resolving #255 (at least the `.eslintrc` part), before the rules in other repos start diverging again? :)

src/components/TabSelect/TabSelect.js
guyfedwards: agreed, added the reasoning also

PR #312, opened 2018-07-20
Extend Dialog component props

fbarl:
Brings the `Dialog` component closer to / easier to use for https://github.com/weaveworks/service-ui/issues/2813#issuecomment-405962739 (inviting team members to Weave Cloud).

### DIFF
--- a/src/components/Dialog/Dialog.js
+++ b/src/components/Dialog/Dialog.js
@@ -1,4 +1,3 @@
-// eslint-disable
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/components/Dialog/Dialog.js
```diff
@@ -106,86 +121,83 @@ const Actions = styled.div`
...
+        </ButtonClose>
+      </Header>
+      <Content>{children}</Content>
+      <Actions>
+        {actions &&
+          actions.map((Action, index) => {
+            if (React.isValidElement(Action)) {
+              /* eslint react/no-array-index-key: 0 */
+              return React.cloneElement(Action, { key: index });
+              /* eslint react/no-array-index-key: 0 */
```
guyfedwards: This needs to be set to 2 again to re-enable the rule as an error

fbarl: Good catch! :+1:

PR #317, opened 2018-07-24
move validation icon inside input

### LINKED ISSUE #315: Remove/re-position icon from validation on Input
Currently we show an icon next to the input when there is an error, this results in either, the input changing width to show the icon or, the input having to allow room for where the icon can show.

I would suggest maybe removing the icon or revisiting how we show errors to the user for fields not only in forms, but when used independently as well. Currently the error does not feel particularly tied to the input especially if the icon is removed.

The usecase I came up against is on the signup form where the icon appearing causes a funky alignment:
image

Couple of options are: 
1. remove completely
2. move the error icon inside the input and pad the input. This also allows for a success green tick to be added on valid if desired
image

3. move icon with validation message and tie them together something like:
image

guyfedwards:
-  validation icon caused change of dimensions and alignment when appearing
  next to the input. This commit moves the icon inside the inputa nd adds a
  padding so the content doesn't go underneath the icon.

- add proptypes, add test, update examples input

fixes #315 
The change to `input` being `width: 100%` will have an impact on existing uses of `Input`. I could only find one instance in service-ui where the inputs were being used without width being explicitly set, in the cloudwatch instructions. I have addressed that in https://github.com/weaveworks/service-ui/pull/2932

image

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -121,7 +121,7 @@ class Input extends React.Component {
         <label htmlFor={id}>{label}</label>
         <InputWrapper valid={valid}>
           {React.createElement(textarea ? 'textarea' : 'input', {
-            ...omit(inputProps, 'autoSelectText'),
+            ...omit(inputProps, 'autoSelectText', 'focus'),
             ref: elem => {
               this.input = elem;
               inputRef(elem);
@@ -132,7 +132,7 @@ class Input extends React.Component {
         <ValidationMessage
           remove={hideValidationMessage}
           valid={valid}
-          visible={message}
+          visible={message && !valid}
         >
           {message}
         </ValidationMessage>
@@ -142,6 +142,14 @@ class Input extends React.Component {
 }
 
 Input.propTypes = {
+  /*
+   * Select the text inside the input. Requires focus attribute to be true
+   */
+  autoSelectText: PropTypes.bool,
+  /*
+   * Focus the input on render
+   */
+  focus: PropTypes.bool,
   /*
    * Remove the validation message from the DOM
    */
@@ -174,6 +182,8 @@ Input.propTypes = {
 };
 
 Input.defaultProps = {
+  autoSelectText: false,
+  focus: false,
   hideValidationMessage: false,
   inputRef: noop,
   label: '',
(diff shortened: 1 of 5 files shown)


src/components/Input/Input.js
```diff
@@ -127,7 +132,7 @@ class Input extends React.Component {
         <ValidationMessage
           remove={hideValidationMessage}
           valid={valid}
-          visible={message}
+          visible={message && !valid}
```
fbarl: Was surprised to find out it wasn't the case yet - good one!

src/components/Input/example.js
```diff
@@ -1,34 +1,48 @@
...
         <Input label="Email" value="ron@hogwarts.edu" onChange={onChange} />
-      </Example>
-      <Example>
+      </StyledExample>
+      <StyledExample>
         <Info>With validation</Info>
         <Input
           label="Email"
-          value="invalid-email"
+          value="invalid-email-that is really long"
```
fbarl: Would be nice to actually make this one editable so that we can test the icon padding interactively (also, it's less confusing when it's editable).

bia:
Looks good, but I would make sure that a partial letter is not being displayed : 

43152415-6321ae24-8f66-11e8-9393-841d16327dc0

vs

43152415-6321ae24-8f66-11e8-9393-841d16327dc0 copy

Have you considered adding a grey ellipsis?

guyfedwards:
The partial letter would change based on what was typed and the width of the input so can't really stop that from happening sometimes. 

Character limit could be added but I'm not really sure what benefit that would add, and in fact could hinder the user, how long is an email for example?

bia:
Right when the text input field is in focus, there should be no character limit (or maybe like one of 256 characters, the max fro email) and even partial letters would be displayed as the person selects and navigates the field.

When they click outside of the input field, we limit the text displayed to 28 characters, or 25 plus ellipsis.

guyfedwards:
Ah gotcha, so limit would be purely visual, misunderstood initially. 

Again, we would need to know the width of the input and how many characters that could fit in order to enforce a reasonable limit, slightly more tricky but certainly do able. The issue comes really from a variable width font but could probably cater for that with a buffer to account for a reasonable combination of letters.

Or are you suggesting that we would have 25 chars no matter the width eg:
image

This I can see being slightly jarring to the user when clicking and suddenly a long text appearing

bia:
>Or are you suggesting that we would have 25 chars no matter the width eg:

Certainly not - it's purely visual - and I just counted the characters in the last example.

PR #320, opened 2018-07-25
Reduce soft border radius

### LINKED ISSUE #319: Reduce soft border radius from 4px to 2px
Also add it to some other components for UI elements consistency:

* [x] `Button`
* [x] `TimeTravel`
* [x] `MatchedText`

fbarl:
Resolves #319.


bia:
Subtle and classy!
Some notes : 
- dialogue box which pops up is missing rounded corners
- I think we should add roundrects to the legend of the graph colors to match the box which contains them

screen shot 2018-07-25 at 11 58 00

PR #327, opened 2018-07-27
Optimize render cycles of GraphNode and GraphEdge components

fbarl:
Optimization prerequisites for using these in Scope - https://github.com/weaveworks/scope/pull/3262.

### DIFF
--- a/src/components/GraphNode/_GraphNodeStatic.js
+++ b/src/components/GraphNode/_GraphNodeStatic.js
@@ -184,15 +184,26 @@ class GraphNodeStatic extends React.PureComponent {
     );
   }
 
+  handleMouseEnter = ev => {
+    this.props.onMouseEnter(this.props.id, ev);
+  };
+
+  handleMouseLeave = ev => {
+    this.props.onMouseLeave(this.props.id, ev);
+  };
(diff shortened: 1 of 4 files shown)


src/components/GraphNode/_GraphNodeStatic.js
```diff
@@ -0,0 +1,241 @@
...
+
+  render() {
+    const { id } = this.props;
+    const Shape = shapes[this.props.shape];
+
+    return (
+      <GraphNodeWrapper
+        onMouseEnter={ev => this.props.onMouseEnter(id, ev)}
+        onMouseLeave={ev => this.props.onMouseLeave(id, ev)}
+        onClick={ev => this.props.onClick(id, ev)}
```
guyfedwards: I don't know exactly what performance symptoms are being seen but if there is a large number of nodes, I have had problems in the past with garbage collection happening during animations and causing lag. 

One way to lower the amount of garbage collection that occurs is to reduce the amount of inline functions in render as a new function object is recreated every render, more objects === more garbage collection. 

Might not be helpful in this instance but something to think about if there are still janks in animations.

Issue #324, opened 2018-07-26
RFC: Remove padding and margins from Input


guyfedwards:
Currently `Input` has padding and margins on the input itself which are really implementation details, and shouldn't be handled by the `Input` itself.
As a result I find myself manually removing/changing the margins/paddings in a lot of circumstances.

A `margin-bottom` for spacing makes sense, but having a top as well and relying on collapsing margins when the label is not there doesn't seem like the cleanest solution.

We want to have an `Input` component that can be used universally without impacting its surroundings and with minimal overrides.

I am unsure of why said padding/margins are all there so want to open the discussion around this.

![image](https://user-images.githubusercontent.com/1794071/43262628-bca1621e-90d8-11e8-9b49-a7ef3b33b7cf.png)
![image](https://user-images.githubusercontent.com/1794071/43262866-788a29a2-90d9-11e8-8448-887431111a39.png)

guyfedwards:
Largely vice-versa, I have had to remove the padding and leave the margin. 

For example, the `input` has a `margin-bottom: 8px` which I imagine is there to space between the `input` itself and a validation message that might show up. 
I have an implementation where I am trying to add a message below the input component, which should be the consistent `8px` below, but I have both the margin and the padding, giving a total spacing of 18px and thus I must override to make it consistent with the 8px used in other components.
![image](https://user-images.githubusercontent.com/1794071/43312289-b46c471a-9184-11e8-9105-9a7b4d601f9f.png)

I suppose the issue may be more around when and how we apply the padding, as opposed to having it at all which, as you mention and I agree, ensures consistency. The margin below the input for example, should really be on the validation message so that it is only present when the validation message is showing. 

The `right-padding` on an `input` is there to consistently space between `input`s I would assume? So it might make more sense to move it from being always there, to being `left-padding` only on inputs that are next to other inputs. This way when you want to have a full width `input` you don't need to remove the right padding.

I have been using the `input`s for a short space of time and only in two situations but in both I have found myself needing to tweak and remove padding/margins so wanted to open the discussion to find out more around what the spacing/layout rules are and how the input/form components should interact with each other, then we can look to tweak the implementation a little to result in less custom code being written for these components.

guyfedwards:
A follow up question:
Is there somewhere I can see design docs for this kind of stuff? Padding/layouts/componetns etc?

bowenli:
[...]

PR #329, opened 2018-07-27
tweak rules around validation display

guyfedwards:
Only show the message if there is a message value *and* the field
is invalid.
Move the margin for spacing between input and message to the message
itself so the space is not there when the message is not.

related #324

Spacing before has message margin even though there is no message to show:
image

input_after

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -32,7 +32,6 @@ const InputWrapper = styled.div`
   display: flex;
   align-items: center;
   margin-top: 10px;
-  margin-bottom: 8px;
   height: 36px;
 
   input {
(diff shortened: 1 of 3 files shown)


bowenli:
Removing the margin and adding the error text on the fly can have unintended consequences. If I understand correctly, what will happen is all content after this field will get pushed down. This can potentially break the flow of the page if it is not accounted for properly. Do you have a list of where we are using this input field? I don't think we currently use it in a way that would break, but it would be good to check.

guyfedwards:
So I can compile a list of Input uses but the current implementation works the same - adding/removing the error message - so the only difference is the initial margin that shows without the error message, the pushing of other content is already happening as shown below:
input_error_margin_before

guyfedwards:
So the list looks like this:
```
team-selector-dropdown
flux-release-notebook-selector
team-instances
notifications/page
settings-form/input
aws-cloud-watch/setup-instructions
```
Haven't got round to seeing the exact pages these are shown on but I don't this PR should cause any issues based on the current implementation also shows/hides the error message. @bowenli thoughts?

bowenli:
@guyfedwards I see what you are saying.

Separately, after this change, someone will have to go fix some of the existing implementations that depend on the margin for spacing, eg. on that cloudwatch page.

guyfedwards:
@bowen have opened https://github.com/weaveworks/service-ui/pull/2992 to address the cloudwatch inputs. 

The issue there was actually that they were using a custom implementation, padding was removed and label was separate to `Input`. So after this change and changing the cloudwatch inputs to use the input 'as-is' the layout is preserved. Thanks for pointing that out 🎉

PR #332, opened 2018-07-31
Migrate to CircleCI 2.0

### LINKED ISSUE #328: Move to CircleCI 2.0
From any of our jobs in Circle CI:

image

We need to do it soon!

fbarl:
Resolves #328.

Followed these instructions for migration: https://circleci.com/docs/2.0/migrating-from-1-2/

The average running time seems to improve from over 3 minutes to about 1.5 mintues :)


fbarl:
> Nice. Might be able to get it even quicker using mhart/alpine-node:8.4.0 as image size is ~1/10th the size.

Cool - I switched to it, it does seem to spin up a bit faster. The problem with it was that saving and restoring `yarn` cache didn't quite work, but I've removed that part to simplify the job as it was only saving us a few seconds per run.

PR #336, opened 2018-08-06
Add DOM ref to GraphNode and GraphEdge components

fbarl:
Needed for https://github.com/weaveworks/scope/pull/3262.


src/components/GraphNode/GraphNode.js
```diff
@@ -165,6 +172,7 @@ GraphNode.defaultProps = {
   cursorType: 'pointer',
   renderPrependedInfo: noop,
   renderAppendedInfo: noop,
+  graphNodeRef: noop,
```
foot: should it have a default value? `undefined` is what react usually expects if you're not doing anything w/ the ref.

PR #340, opened 2018-08-08
Support linkable deployment annotations

### LINKED ISSUE #339: Make deployment annotations linkable
Should be done for both `PromGraph` and `TimeTravel` components.

fbarl:
Resolves #339.

### DIFF
--- a/src/components/_DeploymentAnnotation/DeploymentAnnotation.js
+++ b/src/components/_DeploymentAnnotation/DeploymentAnnotation.js
@@ -1,4 +1,3 @@
-/* eslint-disable import/no-extraneous-dependencies */
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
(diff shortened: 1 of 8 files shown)


src/components/_DeploymentAnnotation/DeploymentAnnotation.js
```diff
@@ -64,14 +64,17 @@ const InfoLine = styled.span`
   overflow: hidden;
 `;
 
-class DeploymentAnnotations extends React.PureComponent {
-  constructor(props) {
-    super(props);
+const NoLink = styled.span`
+  cursor: default;
```
guyfedwards: Is this needed, `span` should have default anyway?

fbarl: The default value is normally `inherit` I think and in case of `TimeTravel`, the parent element has cursor `move` - without this line, we can never be sure it's properly enforced.

PR #349, opened 2018-08-15
Adds onFilterSelect handler to allow more flexibility in filter combination.

foot:
- E.g. to be able to clear the input and replace the contents of the search box w/ the filter. ala github

### DIFF
--- a/src/components/Search/Search.js
+++ b/src/components/Search/Search.js
@@ -75,7 +75,7 @@ const Styled = component => styled(component)`
 `;
 
 class Search extends React.PureComponent {
-  addSearchTerm = (ev, value) => {
+  addSearchTerm = value => {
     let nextPinnedTerms = this.props.pinnedTerms;
     // only push unique values
     if (!includes(nextPinnedTerms, value)) {
@@ -93,7 +93,7 @@ class Search extends React.PureComponent {
   handleInputKeyPress = ev => {
     if (ev.key === 'Enter' && this.props.query.length > 0) {
       ev.preventDefault();
-      this.addSearchTerm(ev, this.props.query);
+      this.addSearchTerm(this.props.query);
     } else if (ev.key === 'Backspace' && this.props.query === '') {
       ev.preventDefault();
       const term = last(this.props.pinnedTerms);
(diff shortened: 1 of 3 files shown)


src/components/Search/Search.js
```diff
@@ -117,8 +117,7 @@ class Search extends React.PureComponent {
   };
 
   handleFilterChange = (ev, value) => {
-    this.input.focus();
```
fbarl: Maybe we can keep this one as I think we'd want it in all cases and might be hard to do it externally without the `input` DOM reference.

src/components/Search/Search.js
```diff
@@ -117,8 +117,7 @@ class Search extends React.PureComponent {
...
-    this.addSearchTerm(ev, value);
```
fbarl: I just realized `addSearchTerm` doesn't use `ev`, maybe you can remove it from there?

src/components/Search/Search.js
```diff
@@ -204,6 +208,7 @@ Search.defaultProps = {
   placeholder: '',
   filters: [],
   onPin: noop,
+  onFilterSelect: noop,
```
fbarl: Wonder if we could make `addSearchTerm` a default callback here. Or else I'd make it a required prop, wouldn't make sense to let the user go without specifying this callback, otherwise the filters would just not work ;)

src/components/Search/example.js
```diff
@@ -80,11 +89,11 @@ class SearchExample extends React.Component {
               })
             }
             filters={[
+              { value: 'red', label: 'Red' },
               {
-                value: 'red',
-                label: 'Red Red Red Red Red Red Red Red Red Red Red Red Red',
+                value: 'blue',
+                label: 'Blublublublublublublublublublublu',
```
fbarl: Excellent choice :ok_hand:

src/components/Search/Search.js
foot: Interesting thought! Replacing default behaviour sounds confusing, I wonder if you can declare dependent proptypes...

src/components/Search/Search.js
foot: We can try that out.

PR #354, opened 2018-08-20
Merge success and warning status colors with accent colors

### LINKED ISSUE #148: Remove semantic sorting of colours
Currently colours are sorted by
- primary
- accent
- status
- and other labels like 'legacy.'

( context : @fbarl and I plan to consolidate all colors used in service-ui, scope, and ui-components into the single theme/weave.js file. }

When all of the colors are together in one file, we might consider a more direct grouping, e.g. : 
- greys
- greens
- blues
- reds

or something else?
cc @jpellizzari

bia:
Resolves #148 and #178 - see https://github.com/weaveworks/ui-components/issues/148#issuecomment-413875368

~Should I be changing the version here too?~

### DIFF
--- a/styleguide/Colors.js
+++ b/styleguide/Colors.js
@@ -165,11 +165,12 @@ class Colors extends React.Component {
         <Text large bold>
           Accent Colors
         </Text>
-        <Row>{this.renderSwatches(theme.colors, /^(orange|blue)[0-9]*$/)}</Row>
-        <Text large bold>
-          Status Colors
-        </Text>
-        <Row>{this.renderSwatches(theme.colors.status)}</Row>
+        <Row>
+          {this.renderSwatches(
+            theme.colors,
+            /^(orange|blue|green|yellow)[0-9]*$/
+          )}
+        </Row>
         <Text large bold>
           Neutral Colors
         </Text>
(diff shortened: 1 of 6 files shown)


foot:
> Should I be changing the version here too?

No, we do a version bump after a merge into master. Would be nice to automate that..

Will take a look.

foot:
Nice, looks good!

wrt keeping service-ui in sync w/ these changes. We should prep a PR in service-ui that migrates it from `status.success/warning` to the new colors so we can merge that in at the same time as this.

styleguide/Colors.js
```diff
@@ -165,11 +165,7 @@ class Colors extends React.Component {
         <Text large bold>
           Accent Colors
         </Text>
-        <Row>{this.renderSwatches(theme.colors, /^(orange|blue)[0-9]*$/)}</Row>
-        <Text large bold>
-          Status Colors
-        </Text>
-        <Row>{this.renderSwatches(theme.colors.status)}</Row>
+        <Row>{this.renderSwatches(theme.colors, /^(orange|blue|green|yellow)[0-9]*$/)}</Row>
```
foot: Prettier is complaining about this line, tooo long

bia:
(yes, that's what I'm doing now, but I was in a meeting)

styleguide/Colors.js
bia: Ops! Prettier now.

foot:
Merge this guy into master.

Then run yarn publish on master. Do you remember if you've got access to aws etc to run that command? I don't actually. I should really do that...

Then once that completes we head back to service-ui branch and change the verison in package.json and merge that when all the checks are ✅

PR #358, opened 2018-08-24
only render children when the dialog is active

### LINKED ISSUE #357: Dialog should not render children until it is visible
Currently the way dialog works means that even though dialog is hidden, the child component is still rendered. Therefore any computation is done regardless of whether the dialog is active or not. 

Ideally the dialog not be added to the DOM until it is needed, but for ease of maintaining current transitions adding conditional for `children` should suffice.

```
{active && children}
```

guyfedwards:
Due to being hidden with css only, and still added to the DOM, any child
component is instantiated when the containing page is rendered.
If you have  a Dialog child with any heavy computation, or any at all
for that matter, you would be doing it even though the dialog may never
be opened.
This commit changes that by only rendering the children when the dialog is
active.

fixes #357

### DIFF
--- a/src/components/Dialog/Dialog.test.js
+++ b/src/components/Dialog/Dialog.test.js
@@ -1,5 +1,8 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
+
+import 'jest-styled-components';
+import { withTheme } from '../../utils/theme';
 
 import Dialog from './Dialog';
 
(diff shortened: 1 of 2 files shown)


src/components/Dialog/Dialog.test.js
```diff
@@ -38,11 +41,34 @@ describe('<Dialog />', () => {
...
-    const wrapper = shallow(<Dialog />);
+    const wrapper = shallow(<Dialog active />);
     expect(wrapper.find('i.fa.fa-close').length).toBe(1);
     wrapper.setProps({
       hideClose: true,
     });
     expect(wrapper.find('i.fa.fa-close').length).toBe(0);
   });
+
+  it('should only render child when active', () => {
```
foot: ⭐️

PR #371, opened 2018-09-07
add checkbox toggle bg for new color styleguide

guyfedwards:
Adds a checkbox to toggle whether the background wrapping all the color
swatches should be updated to the new color entered in the input field.

Makes it easier to see which color is the closest as opposed to having the
'new color' swatch at the end of each row


styleguide/Colors.js
```diff
@@ -157,42 +174,48 @@ class Colors extends React.Component {
             value={this.state.newColor}
             onChange={this.handleChange}
           />
+          <label style={{ marginLeft: '10px' }}>
+            <input type="checkbox" onChange={this.toggleBg} />
+            Toggle bg
```
fbarl: `Toggle background`? :P I don't think I'd know what `bg` refers to from the UI, but then again I'd probably just click it :)

Issue #372, opened 2018-09-07
Update to react@v16.5.0


guyfedwards:
Many fixes and improvements, main addition being support for 'React DevTools Profiler'

facebook/react#13058

Will inevitably mean updating some/all related react-* packages

PR #378, opened 2018-09-14
update jest, jest-styled-components, babel-jest

guyfedwards:
jest@23.6.0
jest-styled-componetns@6.2.0
babel-jest@23.6.0

prerequisite for #372


src/components/Search/Search.js
```diff
@@ -221,7 +221,7 @@ Search.propTypes = {
    * Handler that runs when an item from the search filter dropdown is selected
    * Returns the selected filter value.
    */
-  onFilterSelect: PropTypes.func,
+  onFilterSelect: PropTypes.func.isRequired,
```
foot: `onFilterSelect` is only required if `filters` is provided. I think I tried to write a dependant type thing but it got messy[0] and so fell back to the run-time hack in the defaultProps below.

I think it would be nice to have onFilterSelect optional somehow, including https://github.com/jamiebuilds/react-required-if could be an okay solution, or making our own similar wrapper and chucking it in utils could work too! WDYT?

[0] https://stackoverflow.com/questions/42299335/react-props-set-isrequired-on-a-prop-if-another-prop-is-null-empty

guyfedwards: Yep sounds good. Added `react-required-if` as no point re-implementing that 👌

src/components/Search/Search.js
```diff
@@ -221,7 +222,7 @@ Search.propTypes = {
    * Handler that runs when an item from the search filter dropdown is selected
    * Returns the selected filter value.
    */
-  onFilterSelect: PropTypes.func.isRequired,
+  onFilterSelect: requiredIf(PropTypes.func, props => props.filters),
```
foot: I'm not sure if props here has been merged w/ defaultProps? (If so then this should be filters.length > 0 / or we change defaultProps.filters to undefined.)

guyfedwards: Great point, `filters.length > 0` should suffice as you can't select a filter if there aren't any

Issue #380, opened 2018-09-18
upgrade to webpack 4


guyfedwards:
Prequisite for babel 7 and webpack 1,2,3 are unsupported

PR #381, opened 2018-09-19
update webpack@4, loaders

guyfedwards:
brings us up to date with webpack and associted plugins

fixes https://github.com/weaveworks/ui-components/issues/380


bin/dev-server.js
```diff
@@ -11,4 +12,5 @@ new WebpackDevServer(webpack(config), {
   if (err) {
     throw err;
   }
+  console.log('Started server at http://localhost:8080');
```
foot: 👌

package.json
```diff
@@ -89,31 +90,28 @@
...
     "eslint-plugin-react": "7.6.1",
     "expect": "1.20.2",
-    "extract-text-webpack-plugin": "^1.0.1",
+    "extract-text-webpack-plugin": "^3.0.2",
     "faker": "^4.1.0",
-    "file-loader": "^0.9.0",
+    "file-loader": "^2.0.0",
     "font-awesome": "4.7.0",
-    "font-awesome-webpack": "0.0.4",
+    "font-awesome-webpack": "https://github.com/jarecsni/font-awesome-webpack#440af2a2efe7ba1779d039556f04538e57bad4bb",
```
foot: Maybe a quick comment about why we're not using the default repo.

guyfedwards: No comments in JSON 😞 

We are using this fork as the original repo is not up to date for webpack 4.

I also wonder if we even need this going forward or if there are alternative ways to include the fonts without a dependency

PR #390, opened 2018-10-20
Fixed styling for Alert component

sushma-priyadharssini:
Fixes [#356](https://github.com/weaveworks/ui-components/issues/356)

PR #395, opened 2018-10-23
Removing default alert style

bia:
See [ui-components/pull/#390comment](https://github.com/weaveworks/ui-components/pull/390#pullrequestreview-167414511).
Removing default alert style and changing unspecified styles to info.

### DIFF
--- a/src/components/Alert/example.js
+++ b/src/components/Alert/example.js
@@ -6,11 +6,9 @@ import Button from '../Button';
 import Alert from '.';
 
 const initial = {
-  default: true,
   error: true,
   info: true,
-  multi: true,
-  multiTitle: true,
+  title: true,
   success: true,
   warning: true,
 };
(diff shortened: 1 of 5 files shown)


src/components/Alert/example.js
```diff
@@ -65,31 +60,6 @@ export default class AlertExample extends React.Component {
             Error: Wow you really screwed this up...
           </Alert>
         </Example>
-        <Example>
```
guyfedwards: Why are the multiline and multiline+title examples being removed? Seems separate to this PR?

guyfedwards: I couldn't find anywhere explicitly using the multiline feature so maybe fine to remove. I also couldn't find anywhere we use the title but I think having an example showing the title is worthwhile. 

Could maybe remove just the multiline and change the title one to be a `warning` or `error` to align with the warning icon?

rndstr: > and didn't want people to take that as an example of something to do

Are we saying multiline alerts should not be used?

+1 on title/icon.

src/theme/misc.js
```diff
@@ -1,7 +1,7 @@
 import { colors } from './colors';
 
 export const overlayIconSize = '300px';
-export const textColor = colors.black;
+export const textColor = colors.white;
```
guyfedwards: This seems like a fairly large change? The snapshot above suggests that the `Text` component now has `color: hsl(0,0,100);` which seems like an issue?

bia: My mistake! Didn’t realise how big of a change that would be.

bia:
>Are we saying multiline alerts should not be used?

Yes, if possible.

>I think the Default example could also be removed, the fact that we default to Info style should be obvious from the table above.

It was already - screenshot was just not up to date.
Now it looks like
screenshot 2018-10-25 at 15 17 29

PR #399, opened 2018-10-30
add max-height for dropdown panel

guyfedwards:
very long lists go off the bottom of the screen and so their values
cannot be accesssed. This commit sets a max-height of 10 items, and
then scroll.


fbarl:
Perhaps adding an example would be good?

guyfedwards:
Added more items to the default as don't think it warrants a dedicated example for default behaviour

PR #402, opened 2018-10-31
only render labal if there is a label

guyfedwards:
This commit stops adding <label> to the DOM, when there is no
label prop provided. It also removes the margin - used to add space
between the label and input - when there is no label prop provided.

Removed margin reset that was no longer needed on Search component

### DIFF
--- a/src/components/Input/Input.test.js
+++ b/src/components/Input/Input.test.js
@@ -60,6 +60,22 @@ describe('<Input />', () => {
     );
   });
 
+  it('should only render <label> if label prop is provided', () => {
+    props = {
+      valid: false,
+      message: 'Derp!',
+    };
+    wrapper = mount(withTheme(<Input {...props} />));
+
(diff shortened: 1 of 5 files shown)


src/components/Input/Input.test.js
```diff
@@ -60,6 +60,22 @@ describe('<Input />', () => {
...
+    wrapper = mount(withTheme(<Input {...props} />));
+
+    expect(wrapper.find('label').length).toBe(0);
+
+    wrapper = mount(
+      withTheme(<Input {...{ ...props, label: 'Woohoo, label time!' }} />)
+    );
+
+    expect(wrapper.find('label').length).toBe(1);
+  });
```
fbarl: Maybe also add an example with no label for comparison?

guyfedwards: Thats what the couple of lines above are doing, checks that without `label`, `length === 0`.

Issue #223, opened 2018-05-17
Align singular/pluralisations of theme css namespaces (borderRadius.soft vs fontFamilies.monospace)


foot:
Singular
- borderRadius 
- boxShadow 

Plural:
- fontSizes
- fontFamilies

PR #403, opened 2018-11-01
[RFC] - add selectors for grabbing theme values

guyfedwards:
Composable selectors for grabbing values from props and specifically
theme in styled components.

Changes this:
`background-color: ${props => props.theme.colors.orange};`
into
`background-color: ${color('orange')}`

After implementing this I discovered the `fromAtoms` util and generally the `Atoms`
concept. @fbarl could you give me a breakdown about the thought behind that?

### DIFF
--- a/src/theme/selectors-test.js
+++ b/src/theme/selectors-test.js
@@ -1,0 +1,29 @@
+import { selector, color, fontSize } from './selectors';
+
+import theme from '.';
+
+const props = { theme };
+
+describe('ThemeSelectors', () => {
+  describe('fontSizes', () => {
+    it('should retrieve a fontSize', () => {
+      expect(typeof fontSize).toBe('function');
+      expect(fontSize('small')(props)).toBe(10);
+    });
+  });
+
+  describe('colors', () => {
+    it('should retrieve a color', () => {
+      expect(typeof color).toBe('function');
+      expect(color('grey200')(props)).toBe('#333');
+    });
+  });
+
+  describe('custom selector', () => {
+    it('should return a function ', () => {
+      const wongoBongo = selector('wongo.bongo');
+      expect(typeof wongoBongo).toBe('function');
+      expect(wongoBongo('jibberish')).toBe('boo');
+    });
+  });
+});
--- a/src/theme/selectors.js
+++ b/src/theme/selectors.js
@@ -1,0 +1,20 @@
+import { get, isObject } from 'lodash';
+
+// `selector()` allows you to create closures over as
+// many levels as you wish, each time returning another function.
+// This lets you compose selectors and when they are passed an object
+// finally select that value
+export const selector = (propsOrSelector, prevArgs = []) => {
+  if (isObject(propsOrSelector)) {
+    return get(propsOrSelector, prevArgs);
+  }
+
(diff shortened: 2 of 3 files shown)


foot:
This is pretty slick :). Can you `type` it 😈 ?

I like the terseness, the selector fn itself takes a second to read. I wonder what alternate syntaxes look like that have a simpler base.

1. `font-size: ${fontSize('small')}` - proposed here.
2. `font-size: ${themeGet('fontSize.small')}` - a `get()` wrapper
3. `font-size: ${fontSize.small}` <- w/ a proxy / getter / setter? this is probably not simpler to understand..

- Haven't look at the atom stuff in a while.
- Flicked through https://www.styled-components.com/ecosystem but there's not much eye catching.

guyfedwards:
Yea the idea with the selector is to be allow any composed functions to be created and not just limit it to theme ¯\\_(ツ)_/¯

src/theme/selectors-test.js
```diff
@@ -0,0 +1,41 @@
...
+    colors: {
+      grey200: '#333',
+    },
+    wongo: {
+      bongo: {
+        jibberish: 'boo',
+      },
+    },
+  },
+};
```
fbarl: Why not use the real theme file here?

src/theme/selectors.js
```diff
@@ -0,0 +1,16 @@
...
+
+  return value => selector(value, [...prevArgs, propsOrSelector]);
+};
+
+const themeSelector = selector('theme');
+
+export const boxShadow = themeSelector('boxShadow');
+export const borderRadius = themeSelector('borderRadius');
+export const color = themeSelector('colors');
+export const fontSize = themeSelector('fontSizes');
```
fbarl: :+1: for plural to singular which partly addressed #223

src/theme/selectors.js
```diff
@@ -0,0 +1,16 @@
+import { get, isObject } from 'lodash';
+
+export const selector = (propsOrSelector, prevArgs = []) => {
+  if (isObject(propsOrSelector)) {
+    return get(propsOrSelector, prevArgs);
+  }
+
+  return value => selector(value, [...prevArgs, propsOrSelector]);
+};
```
fbarl: From what I can see with how this function is currently used, we could simply have:

```
export const selector = key => (
  obj => get(obj, key)
);
```

Or is there another use case you have in mind that is currently not expressed here?

fbarl:
> After implementing this I discovered the `fromAtoms` util and generally the Atoms
concept. @fbarl could you give me a breakdown about the thought behind that?

I've personally never used it, but I remember @jpellizzari put it in place for one of our first components, not sure what we want to do with it - I think the concept is valuable for some cases, but not quite sure we want to keep it global (vs. spread it between component files).

> I wonder what alternate syntaxes look like that have a simpler base...

All the suggestions seem like an improvement but I think my vote still goes for the first suggestion :)

src/theme/selectors.js
fbarl: Ah yes, I see.. Cool, in that case, I'd just copy your comment in the code to make it clear what the use case is and then it'd be ok to merge!

PR #410, opened 2018-11-09
Upgrade to font-awesome 5

foot:
- fa5 has mixed prefixes now. not just `.fa`. (also `.fas`, `.fab`, `.far`), so use `i { }` to restyle things rather than `.fa { }`
- Check that icons looks similar! Had to resize some things here and there, the base icon feels a tiny bit bigger in fa5.
- Included as a peerDep, fa5 to be provided by scope/service-ui.


package.json
```diff
@@ -105,7 +104,7 @@
     "marked": "0.3.12",
     "moment": "2.21.0",
     "node-sass": "4.5.3",
-    "prettier": "^1.12.1",
+    "prettier": "1.10.2",
```
guyfedwards: Is this intentional?

PR #414, opened 2018-11-13
perf: only calculate time and value scales on new data

guyfedwards:
This commit changes PrometheusGraph to only calculate timeScale and valueScale
when there is new data, not every render.
This solves performance bottleneck when 2000+ datapoints and 16graphs on a
page at once.

Profile CPU usage during a release:
Before
screenshot 2018-11-13 15 36 41
After
screenshot 2018-11-13 15 36 49

Time spent in function on canary release page:
screenshot 2018-11-13 14 52 57

### DIFF
--- a/src/components/PrometheusGraph/PrometheusGraph.js
+++ b/src/components/PrometheusGraph/PrometheusGraph.js
@@ -213,6 +213,8 @@ class PrometheusGraph extends React.PureComponent {
       hoverY: null,
       chartWidth: 0,
       chartHeight: 0,
+      timeScale: null,
+      valueScale: null,
     };
   }
 
@@ -252,7 +254,14 @@ class PrometheusGraph extends React.PureComponent {
   };
 
   handleChartResize = ({ chartWidth, chartHeight }) => {
-    this.setState({ chartWidth, chartHeight });
+    this.setState({
+      chartWidth,
+      chartHeight,
+      ...this.prepareTimeAndValueScales(
+        this.props,
+        this.state.visibleMultiSeries
+      ),
+    });
   };
 
   prepareMultiSeries = (props, { selectedLegendKeys } = this.state) => {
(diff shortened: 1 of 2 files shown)


src/components/PrometheusGraph/PrometheusGraph.js
```diff
@@ -351,18 +353,34 @@ class PrometheusGraph extends React.PureComponent {
       })),
     }));
 
-    this.setState({ multiSeries });
+    this.setState({ multiSeries }, () => {
+      // time and value scales use state.multiSeries so must be calculated after
+      // state.multiseries
+      this.prepareTimeAndValueScales(props);
+    });
```
fbarl: Isn't it possible that this part will only be executed in the next render cycle?

src/components/PrometheusGraph/PrometheusGraph.js
```diff
@@ -213,6 +213,8 @@ class PrometheusGraph extends React.PureComponent {
       hoverY: null,
       chartWidth: 0,
       chartHeight: 0,
+      timeScale: {},
+      valueScale: {},
```
fbarl: This should be a function (perhaps a `noop`?), otherwise we get console warnings for type mismatch.

src/components/PrometheusGraph/_Chart.js
```diff
@@ -61,7 +57,7 @@ class Chart extends React.PureComponent {
 
     const cursorValue = valueScale.invert(cursorYOffset);
     const cursorTimestampSec = timeScale.invert(cursorXOffset);
-    const hoverTimestampSec = timestampQuantizer(cursorTimestampSec);
+    const hoverTimestampSec = timestampQuantizer(cursorTimestampSec) || 0;
```
fbarl: I would prefer not to do this - otherwise we say it's ok to show a timestamp `0` which I'd argue is worse than having a warning spit out :)

The proper solution would probably be not to render the timestamp in the tooltip if it's `null` - feel free to do that in `_HoverInfo.js` if you find the warning annoying tho ;)

src/components/PrometheusGraph/PrometheusGraph.js
```diff
@@ -207,12 +207,14 @@ class PrometheusGraph extends React.PureComponent {
       multiSeries: [],
       selectedLegendKeys: props.selectedLegendKeys,
       hoveredLegendKey: null,
-      hoverTimestampSec: null,
+      hoverTimestampSec: 0,
```
fbarl: See the comment below, I'd be happier if we don't do this :)

src/components/PrometheusGraph/PrometheusGraph.js
```diff
@@ -351,18 +360,58 @@ class PrometheusGraph extends React.PureComponent {
...
+
+  prepareTimeAndValueScales = (props, multiseries) => {
+    let chartWidth;
+    let chartHeight;
+
+    if (this.chartRef) {
+      ({
+        width: chartWidth,
+        height: chartHeight,
+      } = this.chartRef.getSvgBoundingRect());
```
fbarl: What's with the `()` wrapper? :D

guyfedwards: Necessary for destructuring into variables that are already declared. I had to google that one 😄

Issue #424, opened 2018-11-22
Apply standard spacing across all components


fbarl:
Follow-up on #400 - probably it makes sense starting with large sizes down to small ones:

* [ ] `64px` xxl
* [ ] `48px` xl
* [ ] `32px` large
* [ ] `24px` ~base~ medium
* [ ] `16px` ~small~ base
* [ ] `12px` small
* [ ] `8px` xs
* [ ] `4px` ~mini~ xxs

bia:
I would go with `xxs` instead of `mini` to avoid confusion.

bia:
We'll address #324 within the scope of this issue.
cc @guyfedwards

rndstr:
> base

my 2c:  is `base` being used anywhere else? this seems more like a `default` to me. `base` tells me that's the value we are basing all the others on which could be something completely different than the 'usual' value to use.

maybe we can name it `medium` instead to go with the sizing names of the others?

rndstr:
@bia that makes totally sense then, thanks for explaining!

Should we also define them in relation to `base` then instead of hardcoding? https://github.com/weaveworks/ui-components/blob/e019f6c486ca968c7bf43ee3047766c6ab15713b/src/theme/spacings.js#L3-L11

PR #425, opened 2018-11-22
Introduce standard spacing to the theme

### LINKED ISSUE #400: Standardise spacing
Now that fonts and colours are quite organised, spacing could still use a little love. 
Airbnb's [design system](https://airbnb.design/building-a-visual-language/) could be a bit of inspiration : 
dls-foundation

We could have a `theme` file which would set some values that are multiples of eachother that we can keep reusing : 
```
export const spacing = {
  xs: '4px',
  mini: '8px',
  small: '16px',
  base: '24px',
  large: '48px',  
  xl: '64px',
};
```
(these values are totally arbitrary and can change as we find appropriate)

cc @fbarl @ngehani

fbarl:
Resolves #400.

### DIFF
--- a/src/theme/spacings.js
+++ b/src/theme/spacings.js
@@ -1,0 +1,22 @@
+import { forEach } from 'lodash';
+
+export const spacing = {
+  none: '0px',
+  xss: '4px',
+  xs: '8px',
+  small: '16px',
+  base: '24px',
+  large: '48px',
+  xl: '64px',
+};
+
+// Collects all theme spacing vars as SCSS vars.
+export function themeSpacingsAsScss() {
+  const themeSpacings = [];
+
+  forEach(spacing, (value, name) => {
+    themeSpacings.push(`$spacing-${name}: ${value}`);
+  });
+
+  return themeSpacings;
+}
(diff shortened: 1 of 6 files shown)


bia:
I'm surprised by how few changes there are here?
Could you also remove `constants.jsx`? I think that's in the scope of this pr.

fbarl:
> I'm surprised by how few changes there are here?

This PR just creates the theme, #424 will apply it to all the components in the repo.

> Could you also remove `constants.jsx`? I think that's in the scope of this pr.

That constant is still being used in `service-ui` repo, but good point - I'll remove it once https://github.com/weaveworks/service-ui/issues/3518 has been addressed :+1:

src/theme/spacings.js
```diff
@@ -0,0 +1,22 @@
+import { forEach } from 'lodash';
+
+export const spacing = {
+  none: '0px',
```
guyfedwards: Do we really need this? and if so would it be better as just 0 seeing as unit is irrelevant and often times you want to explicitly have no border with `border: 0` as opposed to a border of `0px` width. Semantics ¯\\_(ツ)_/¯

fbarl: > Do we really need this?

I put it there just in case, so that if we ever decided to enforce the linting rules, we don't have come back to this just to add the zero :)

> ...would it be better as just 0 seeing as unit...

Agree, will make the change.

Issue #370, opened 2018-09-06
Upgrade styled-components@4 


guyfedwards:
Performance improvements mainly, removal of Component.extend in favour of styled(Component).

https://www.styled-components.com/releases#v4.0.0-beta.0

fbarl:
Resolved by #486.

PR #426, opened 2018-11-26
chore: styled-components@3.4.10

guyfedwards:
related #370


guyfedwards:
@fbarl Still waiting for types to arrive for v4 https://github.com/flow-typed/flow-typed/pull/2933

I tried without success to get 4 working with the types in that PR but seems it is not yet complete. 

Now the `extend`s have been removed the upgrade from 3->4 should be quick.

PR #429, opened 2018-11-27
feat(dropdown): adjusts max-height to show it can be scrolled

### LINKED ISSUE #428: Difficult to see that dropdown can scroll sometimes.
I thought I'd lost some instances from the instances selector the other day before trying to scroll.

- Easy fix would be to make the max-height a multiple of half an item height so that the last item is always chopped in half. A technique that https://react-select.com/home and others probably use.

foot:
- Some OS's don't show scrollbars so we chop the last item in half.

Fixes #428


foot:
Old v. New

screenshot 2018-11-27 at 09 33 48

PR #432, opened 2018-11-27
Standardise component padding and margins

bia:
Continues work after #400 

Here I've corrected #425 to a `16px` base, a [more appropriate model](https://twitter.com/adamwathan/status/1063492067916009473) coming out of discussion with @foot : 
dsjh6wxuwaafslu

Pixel alignment tweaks for symbols and icons will be ignored by this system.
Still to be addressed : 
- Standardising line heights throughout components to match spacing
- Standardise width and height of components?
- Grid component
- In depth breakdown of TimeTravel component
- Scaling components like DotSpinner

I will create new issues for these accordingly.

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -2,6 +2,7 @@ import React from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { map, noop, find, flattenDepth, isArray } from 'lodash';
+import { spacing } from '../../theme/selectors';
 
 const WIDTH = '256px';
 const HEIGHT_NUMBER = 36;
@@ -11,10 +12,9 @@ const Item = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-  padding-right: 28px;
-  padding-left: 12px;
+  padding-right: ${spacing('medium')};
+  padding-left: ${spacing('small')};
   cursor: pointer;
-
   ${props =>
     props.disabled &&
     `
@@ -37,7 +37,7 @@ const Popover = styled.div`
   max-height: ${HEIGHT_NUMBER * 10.5 + 10}px;
   overflow: auto;
   box-sizing: border-box;
-  padding: 6px 0;
+  padding: ${spacing('xxs')} 0;
 `;
 
 const Overlay = styled.div`
@@ -54,7 +54,6 @@ const ItemWrapper = styled(Item)`
   color: ${props =>
     props.selected ? props.theme.colors.blue400 : props.theme.textColor};
   min-height: ${HEIGHT};
-
   &:hover {
     background-color: ${props => props.theme.colors.gray50};
   }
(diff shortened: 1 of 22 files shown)


src/components/Dropdown/Dropdown.js
```diff
@@ -11,10 +13,9 @@ const Item = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-  padding-right: 28px;
-  padding-left: 12px;
+  padding-right: ${spacing('medium')};
```
fbarl: I'm thinking how `spacing('large')` looks better here to me than `medium`, especially when considering search filters:

`large`
image

`medium`
image

Either is fine though, so let it be your call :)

src/components/TabSelect/_TabButton.js
```diff
@@ -14,8 +16,11 @@ const getBgColor = (selected, secondary, theme) => {
 const Styled = component => styled(component)`
   cursor: pointer;
   margin-bottom: -1px;
-  margin-right: 5px;
-  padding: ${props => (props.small ? '5px 10px' : '10px 20px')};
+  margin-right: ${spacing('xxs')};
+  padding-top: ${props => (props.small ? spacing('xs') : spacing('xs'))};
+  padding-bottom: ${props => (props.small ? spacing('xs') : spacing('xs'))};
+  padding-left: ${props => (props.small ? spacing('small') : spacing('base'))};
+  padding-right: ${props => (props.small ? spacing('small') : spacing('base'))};
```
fbarl: Btw. the difference between small and normal tabs is almost erased with these changes, but I think that shouldn't be a big problem:

image

Otherwise I'd consider `medium` instead `base` for horizonal paddings of _normal_ tab selectors..

src/components/TabSelect/_TabButton.js
```diff
@@ -14,8 +16,11 @@ const getBgColor = (selected, secondary, theme) => {
 const Styled = component => styled(component)`
   cursor: pointer;
   margin-bottom: -1px;
-  margin-right: 5px;
-  padding: ${props => (props.small ? '5px 10px' : '10px 20px')};
+  margin-right: ${spacing('xxs')};
+  padding-top: ${props => (props.small ? spacing('xs') : spacing('xs'))};
+  padding-bottom: ${props => (props.small ? spacing('xs') : spacing('xs'))};
```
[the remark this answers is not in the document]
bia: No, `props.small`already existed here to differentiated between the `small` and `default` tab styles. Didn't feel the need to change it.

src/components/Dropdown/Dropdown.js
bia: I thought that `medium` looked more symmetrical, but now I realise that the real issue with the spacing here is that the dropdown content can be so much larger (or smaller) than the width of this filter button. I made #434 to address this.

src/components/TabSelect/_TabButton.js
bia: I also tried
```
  padding-top: ${props => (props.small ? spacing('xs') : spacing('base'))};
  padding-bottom: ${props => (props.small ? spacing('xs') : spacing('base'))};
  padding-left: ${props =>
    props.small ? spacing('small') : spacing('medium')};
  padding-right: ${props =>
    props.small ? spacing('small') : spacing('medium')};
```
It feels like there's too much empty space around the text and I don't really want to make the font larger. I don't really think there is a use case for this and I'd rather too similar than too huge.

screenshot 2018-12-03 at 17 54 45

PR #449, opened 2018-12-11
Bump D3 dependencies

fbarl:
The TimeTravel panning issue appeared again because of older version of D3 being recorded in `yarn.lock`, so I thought it was a good time to bump those deps to get rid of the error.


fbarl:
> d3-select should be a peerdep maybe

That would solve the problem, but would put extra requirements for whoever wants to use our packages... I'd be really happy to learn about a better solution, or to be honest, rather ditch `d3-selection` than think about its versioning.

Perhaps https://github.com/mzabriskie/react-draggable or our own logic would work better :thinking:

PR #463, opened 2019-02-25
feat: allow over-riding internal dropdown components

### LINKED ISSUE #398: Allow more complex data structures in dropdown `item.value`
Currently trying to set an items value as an object doesn't really work, keys become `[Object object]` among other things.

It would be great to be able to pass an items array such as:
```
items: Array<{
  value: {
    [string]: any
  } | string | number,  
  label: React.Node,
  id: string
}>
```

For clarity I think it would benefit from an additional property of `id` rather than trying to be clever with value/label as both of those should be able to be nodes

foot:
- Inspired by https://react-select.com/components

Fixes #398

### DIFF
--- a/docs/js/Example.js
+++ b/docs/js/Example.js
@@ -56,6 +56,10 @@ const TableRow = styled.tr`
     background-color: ${props => props.theme.colors.purple100};
   }
 
+  p {
+    margin: 0;
+  }
+
   td {
     padding: 3px 4px;
   }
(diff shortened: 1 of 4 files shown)


foot:
Hey gang what do you think of this approach?

Options:
- this solutions w/ `Components=`
- `renderItem=`
- more complex data structures as suggested in the linked issue which I found after I made this PR oops.

foot:
> we should try to ensure that we don't end up with lots of different dropdown implementations throughout the app.

Well put, this is the main concern, all of the approaches are vulnerable, this one here perhaps slightly more so as we're replacing the entire ItemWrapper rather than just letting you fill it more flexibly ... 🎲

foot:
I chickened out and moved back to something more conservative as the issue proposes.

PR #469, opened 2019-03-22
workaround for dropdown item label title `[object Object]`

dimitropoulos:
Screenshot_20190322_123747

The above UI bug can happen in a situation where the label passed in is an object since objects are truthy in javascript.  Ideally, Flow should catch this kind of problem but since it doesn't (from what I can tell the type gets lost as it goes through the `lodash/find` on line 200), a type check here will work for now.

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -228,7 +228,9 @@ class Dropdown extends React.Component {
                       this.handleChange(ev, item.value, item.label)
                     }
                     selected={item.value === value}
-                    title={item && typeof item.label === "string" ? item.label : ""}
+                    title={
+                      item && typeof item.label === 'string' ? item.label : ''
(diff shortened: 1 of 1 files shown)


rndstr:
@errordeveloper the problem is that the eksctl within `weaveworks/eksctl:0.1.25` has no version

```
 $ docker run weaveworks/eksctl:0.1.25 eksctl version
[ℹ]  version.Info{BuiltAt:"1553524352", GitCommit:"0.1.25-dirty", GitTag:""}
```

dimitropoulos:
thanks @rndstr - I'll hopefully be setting up my environment for javascript dev soon enough but you saved me from having to do it now :)

fbarl:
ping @dimitropoulos - anything still blocking you from merging this?

dimitropoulos:
no - merging now.

PR #472, opened 2019-04-17
New design of ResourceDial component

### LINKED ISSUE #471: New design of ResourceDial component
@bia has made a design update to the `ResourceDial` component:

> because they take up less space and *give a better sense of full vs current* [like in grafana]

##### Old
image

##### New
image

We should update the component to reflect the new look.

bia:
Resolves #471

Before
Screenshot 2019-04-17 at 16 28 29

After
Screenshot 2019-04-17 at 16 28 46


foot:
Screenshot 2019-04-18 at 11 26 13

Looks great! The only niggle is there's a bit of padding down the bottom now which might make it look a bit funny when placed in a container.

bia:
It's even a little more compact now (I choose `large` padding instead of `xl`), but I think I like it.

Screenshot 2019-04-18 at 11 50 14

PR #474, opened 2019-05-20
Refine dotted GraphNode shapes

bia:
Currently the shapes look rather fuzzy,
Screenshot 2019-05-20 at 15 41 15

so you could say that I made them dashed, rather than dotted.
Screenshot 2019-05-20 at 15 40 25

This is work towards #475


bia:
>I think you were gonna reference an issue at the PR description but the number's missing ;)

Ops! Added.

>the perspective is not quite right and that's definitely something outside of the scope of this PR

Semantically, there's a problem with having one 3D shape among a set of 2D shapes. But this is for another issue.

PR #477, opened 2019-05-24
fix(timeline): enable preventDefault() on onWheel event

fbarl:
Temporary hack solution for https://github.com/weaveworks/service-ui/issues/3750.

Also, use left alignment as default for `GridRow` component (which changes nothing as `flex-start` is already a Flex default attribute) so that `alignContent` is not a required prop anymore.


src/utils/zooming.js
```diff
@@ -12,3 +12,60 @@ function wheelDelta(ev) {
...
+  const EVENTS_TO_MODIFY = [
+    'touchstart',
+    'touchmove',
+    'touchend',
+    'touchcancel',
+    'wheel',
+  ];
+
+  const originalAddEventListener = document.addEventListener.bind();
+  document.addEventListener = (type, listener, options, wantsUntrusted) => {
```
foot: This is a bit scary replacing the global `addEventListener` fn.

Would an approach like we use in scope https://github.com/weaveworks/scope/blob/master/client/app/scripts/components/zoomable-canvas.js#L66 work too?

fbarl: Yup, I tried it and it seems to work - thanks for the tip!

PTAL

PR #481, opened 2019-05-31
fix: tweaks dropdown-item disabled style

foot:
output


src/components/Dropdown/Dropdown.js
```diff
@@ -20,8 +20,8 @@ const Item = styled.div`
     props.disabled &&
     `
     cursor: not-allowed;
-    color: ${props.theme.colors.gray600};
-    background-color: ${props.theme.colors.gray50};
+    color: ${props.theme.colors.gray200};
+    background-color: ${props.theme.colors.white};
```
foot:
```suggestion

```

PR #486, opened 2019-09-04
feat: updates to styled-components 4

foot:
- Upgrades to styled-components-4
- Upgrades dev deps to react-16.9, enzyme, jest etc to latest.

Code transformations for SC4:
- `.attrs` has lifted the fn from each prop up to the top of the attr block: https://www.styled-components.com/docs/api#attrs 
  - old: `{ fontSize: props => props.theme.fontSize }`
  - new: `props => { fontSize: props.theme.fontSize }`
- New global injection: `injectGlobal` -> `createGlobalStyle`
- Use react prop forwarding: `<Component innerRef={} ...` -> `<Component ref={} ...`
- `keyframes` can't be directly interpolated into a string and have to be wrapped in `css()` block
- Minor `.snap` changes: SC now filters things out like "icon" before they hit the dom
- Major `.snap` changes: SC jest-renderer has changed a bit v3 -> v4 and puts more stuff in the props.


dimitropoulos:
(styled-components release notes: https://github.com/styled-components/styled-components/blob/master/CHANGELOG.md)

PR #494, opened 2019-11-27
Dialog title bigger

### LINKED ISSUE #493: Consider making dialog title bigger
#### Current

* Dialog padding `16px 24px 24px`
* Title font size: `16px`
* Title font weight: `bold`

image

#### Suggested

image

* Dialog padding `24px`
* Title font size: ~`24px`~ `22px`
* Title font weight: `normal`

cc @bia

bia:
Fixes #493

### DIFF
--- a/src/components/Dialog/Dialog.js
+++ b/src/components/Dialog/Dialog.js
@@ -47,7 +47,7 @@ const Window = styled.div`
   width: ${props => props.width};
   margin: 0 auto;
   max-width: 768px;
-  padding: ${spacing('base')} ${spacing('medium')} ${spacing('medium')};
+  padding: ${spacing('medium')};
   position: relative;
 `;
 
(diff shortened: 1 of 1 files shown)


src/components/Dialog/Dialog.js
```diff
@@ -58,7 +58,7 @@ const Header = styled.div`
 `;
 
 const Title = styled.span`
-  font-size: ${props => props.theme.fontSizes.normal};
+  font-size: ${props => props.theme.fontSizes.large};
   font-weight: bold;
```
fbarl: Maybe we want to drop this line as well to make it more similar to spec in #493 and also Material UI?

#### Bold

image

#### Normal

image

bia: Thanks @fbarl, I agree!

PR #496, opened 2020-01-26
style: dropdown and input have same style when disabled

### LINKED ISSUE #492: Dropdown and Input have different styles when disabled
These two should look more similar:

image
image

IMO, the former (`Input`) component looks better when disabled so we can just tweak the styles of `Dropdown` component.

skarunakar:
Fixes #492.

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -20,8 +20,8 @@ const Item = styled.div`
     props.disabled &&
     `
     cursor: not-allowed;
-    color: ${props.theme.colors.gray200};
-    background-color: ${props.theme.colors.white};
+    color: ${props.theme.colors.gray600};
+    background-color: ${props.theme.colors.gray50};
   `};
 `;
 
@@ -79,6 +79,8 @@ const SelectedItem = styled(Item)`
   div:last-child {
     margin-left: auto;
   }
+  ${props =>
+    props.disabled && `background-color: ${props.theme.colors.gray50};`};
 `;
 
 const SelectedItemIcon = styled.span`


src/components/Dropdown/Dropdown.js
```diff
@@ -79,6 +79,11 @@ const SelectedItem = styled(Item)`
   div:last-child {
     margin-left: auto;
   }
+    ${props =>
+    props.disabled &&
+    `
+    background-color: ${props.theme.colors.gray50};
+  `};
```
skarunakar: dropdown-disabled

Dropdown disabled with above changes

src/components/Dropdown/Dropdown.js
```diff
@@ -79,6 +79,8 @@ const SelectedItem = styled(Item)`
   div:last-child {
     margin-left: auto;
   }
+  ${props =>
+    props.disabled && `background-color: ${props.theme.colors.gray50};`};
 `;
```
skarunakar: dropdown-disabled
Dropdown disabled with this changes

skarunakar: input-disabled

Input when disabled

src/components/Dropdown/Dropdown.js
```diff
@@ -20,8 +20,8 @@ const Item = styled.div`
     props.disabled &&
     `
     cursor: not-allowed;
-    color: ${props.theme.colors.gray200};
-    background-color: ${props.theme.colors.white};
+    color: ${props.theme.colors.gray600};
```
skarunakar: gray600 is color of placeholder text in input component when disabled. Not sure if it should be rgb(0, 0, 0) or something else

fbarl: `gray600` seems good!

src/components/Dropdown/Dropdown.js
fbarl: Perfect, looks much better IMO :)

bia:
Also looks like a great improvement to me, let’s merge. 👍 
Thank you @skarunakar !

PR #514, opened 2021-03-15
chore: upgrade to styled-components 5

foot:
https://styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v5

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,9 +55,9 @@
     "@fortawesome/fontawesome-free": "^5.0.0",
     "lodash": "^4.17.12",
     "moment": "^2.0.0",
-    "react": "^16.3.0",
-    "react-dom": "^16.3.0",
-    "styled-components": "^4.3.2"
+    "react": "^16.8.0",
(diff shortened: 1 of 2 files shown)


jpellizzari (commented):
@foot I would expect some changes to the `Code` component around the animation errors we were seeing

foot:
> I would expect some changes to the Code component around the animation errors we were seeing

Indeed! I assumed it was sc4/5 incompatibility but it seems quite broken in sc5. I'll TAL.

Also it would be cool to re-build the docs/examples for each PR..

jpellizzari:
> Also it would be cool to re-build the docs/examples for each PR

Converting to netlify should be no problem

> Indeed! I assumed it was sc4/5 incompatibility but it seems quite broken in sc5. I'll TAL.

It only happens `onClick` if I recall correctly.

foot:
Sorry for the premature review request, tests still need some attention.

